PR #22, opened 2016-09-01
[MIG] connector_prestashop

pedrobaeza:
# 
# Odoo PrestaShop Connector

This module connects Odoo and PrestaShop.

PrestaShop (http://www.prestashop.com/) is a popular e-commerce platform
written in PHP/MySQL and published under the Open Software licence v3.0.

This module allows the synchronization of the following objects from PrestaShop
to Odoo:
- Websites
- Stores and languages
- Carriers
- Product categories
- Products
- Combinations of products
- Partner categories
- Customers

Once these objects are synchronised, it will allow the import of sales orders,
together with the related customers.

As an extra feature, you can also export the stock quantities back to
PrestaShop.

If you want to export from Odoo to PrestaShop changes made on the products,
product categories or product images, you need to install
_connector_prestashop_catalog_manager_ module in this same repository.

This connector supports PrestaShop version up to 1.6.11. Maybe later versions
are also supported, but they haven't been tested. It uses the webservices of
PrestaShop.
# Installation

It doesn't require any plug-in in PrestaShop, but requires an extra Python
library in Odoo server side, called prestapyt:

https://github.com/prestapyt/prestapyt/

You can use pip install system to install it

`sudo pip install prestapyt`
# Configuration

To configure this module, you need to set several things in both PrestaShop
and Odoo:
## Steps in PrestaShop
- Go to the control panel (usually at <url>/adminps).
- Login into the system.
- Go to _Advanced Parameters > Web service_
- Add a new entry.
- Generate a new API key that will be needed later.
- Grant all the needed access according your security policy.
## Steps in Odoo
- Go to _Connectors > PrestaShop > Backends_.
- Create a new record for registering a PrestaShop backend. You will bind
  this backend to an specific company and warehouse.
- Define the main URL of the PrestaShop web, and the webservice key you
  got in PrestaShop.
- Define other parameters like the discount and shipping products, or if the
  taxes are included in the price.
- Click on "Synchronize Metadata" button. This will bring the basic shop
  information that you can find on _Websites_ and _Stores_ menus.
- Click on "Synchronize Base Data" button. This will import carriers,
  languages, tax groups and the rest of base data that are needed for the
  proper work.
- Go to _Accounting > Configuration > Taxes > Tax Groups_, and include
  for each of the tax definition imported from PrestaShop, the corresponding
  taxes in Odoo.
- Activate the job runner, checking the connector documentation for setting
  the server correctly for using it in http://odoo-connector.com/guides/jobrunner.html
- Alternatively, if you are not able to activate it, you can enable the
  scheduled job called "Enqueue Jobs".
- Activate the scheduled jobs for importing the records you want:
  - PrestaShop - Export Stock Quantities
  - PrestaShop - Import Carriers
  - PrestaShop - Import Customers and Groups
  - PrestaShop - Import Products and Categories
  - PrestaShop - Import Sales Orders
  - PrestaShop - Import suppliers
  - PrestaShop - Payment methods
# Usage

To use this module, you need to:
- Go to _Connectors > Queue > Jobs_, and check the correct enqueuing of
  the tasks.
- Check on each menu the resulting imported records (Customers, Sales
  Orders...)
# Known issues / Roadmap
- Work with multiple warehouses.
- Tests.


pedrobaeza:
@guewen here it is!

connector_prestashop/delivery.py
```diff
@@ -1,32 +1,11 @@
...
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
elicoidal: You might want to add the copyright holders

connector_prestashop/models/account.py
```diff
@@ -85,16 +72,19 @@ def _find_refund(self, cr, uid, amount, partner_id, context=None):
...
+    openerp_id = fields.Many2one(
+        comodel_name='account.invoice',
+        required=True,
+        ondelete='cascade',
+        string='Invoice',
+    )
+
+    _sql_constraints = [
+        ('prestashop_erp_uniq', 'unique(backend_id, openerp_id)',
+         'A erp record with same ID on PrestaShop already exists.'),
```
elicoidal: An ERP

connector_prestashop/models/delivery.py
```diff
@@ -0,0 +1,60 @@
...
+
+    openerp_id = fields.Many2one(
+        comodel_name='delivery.carrier',
+        string='Delivery carrier',
+        required=True,
+        ondelete='cascade'
+    )
+    id_reference = fields.Integer(
+        string='Id reference',
+        help="In PrestaShop, carriers with the same 'id_reference' are "
```
elicoidal: In Prestashop, carriers can be copied with same 'Reference ID' (only the last copied carrier will be synchronized with the ERP)

connector_prestashop/models/delivery.py
```diff
@@ -0,0 +1,60 @@
...
+             "some copies from the first one id_reference (only the last "
+             "one copied is taken account ; and the only one which "
+             "synchronized with erp)"
+    )
+    name_ext = fields.Char(
+        string='External name',
```
elicoidal: Name in Prestashop

connector_prestashop/models/delivery.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+    active_ext = fields.Boolean(
+        string='External active',
```
elicoidal: Active in Prestashop

connector_prestashop/models/delivery.py
```diff
@@ -0,0 +1,60 @@
...
+        help="... in prestashop",
+    )
+    export_tracking = fields.Boolean(
+        string='Export tracking numbers',
+        help=" ... in prestashop",
```
elicoidal: not clear: export to Prestashop or is it a field in Prestashop

connector_prestashop/models/delivery.py
```diff
@@ -0,0 +1,60 @@
...
+        default=False
+    )
+
+    _sql_constraints = [
+        ('prestashop_erp_uniq', 'unique(backend_id, openerp_id)',
+         'A erp record with same ID on PrestaShop already exists.'),
```
elicoidal: above

connector_prestashop/models/partner.py
```diff
@@ -0,0 +1,190 @@
...
+
+from openerp import models, fields, api
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    prestashop_bind_ids = fields.One2many(
+        comodel_name='prestashop.res.partner',
+        inverse_name='openerp_id',
```
elicoidal: odoo

connector_prestashop/models/prestashop_model.py
```diff
@@ -0,0 +1,561 @@
...
+        dt = pytz.utc.localize(dt)
+        dt = dt.astimezone(timezone)
+        return dt.strftime('%Y-%m-%d %H:%M:%S')
+
+    @api.multi
+    def import_customers_since(self):
+        session = ConnectorSession(
+            self.env.cr, self.env.uid, context=self.env.context)
+        for backend_record in self:
+            # from_string = fields.Datetime.from_string
```
elicoidal: remove dead code

connector_prestashop/models/prestashop_model.py
```diff
@@ -0,0 +1,561 @@
...
+        string='PrestaShop Backend',
+        required=True,
+        ondelete='restrict'
+    )
+    # TODO : do I keep the char like in Magento, or do I put a PrestaShop ?
+    prestashop_id = fields.Integer('ID on PrestaShop')
+
+    _sql_constraints = [
+        ('prestashop_uniq', 'unique(backend_id, prestashop_id)',
+         'A record with same ID on PrestaShop already exists.'),
```
elicoidal: A record with same ID already exists in PrestaShop.

connector_prestashop/models/prestashop_model.py
```diff
@@ -0,0 +1,561 @@
...
+        string='WareHouse',
+        required=True,
+        readonly=True,
+        ondelete='cascade',
+    )
+    # what is the exact purpose of this field?
+    default_category_id = fields.Many2one(
+        comodel_name='product.category',
+        string='Default Product Category',
+        help="The category set on products when?? TODO."
```
elicoidal: to be improved

connector_prestashop/models/prestashop_model.py
```diff
@@ -0,0 +1,561 @@
...
+        "\nOpenERP requires a main category on products for accounting."
```
elicoidal: Odoo

connector_prestashop/models/prestashop_model.py
```diff
@@ -0,0 +1,561 @@
...
+class PrestashopResLang(models.Model):
+    _name = 'prestashop.res.lang'
+    _inherit = 'prestashop.binding'
+    _inherits = {'res.lang': 'openerp_id'}
+
+    openerp_id = fields.Many2one(
+        comodel_name='res.lang',
+        required=True,
+        ondelete='cascade',
+        string='Lang',
```
elicoidal: Language

connector_prestashop/models/prestashop_model.py
```diff
@@ -0,0 +1,561 @@
...
+    ]
+
+
+class ResCurrency(models.Model):
+    _inherit = 'res.currency'
+
+    prestashop_bind_ids = fields.One2many(
+        comodel_name='prestashop.res.currency',
+        inverse_name='openerp_id',
+        string='prestashop Bindings',
```
elicoidal: Prestashop

connector_prestashop/models/product.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+    # TODO FIXME what name give to field present in
+    # prestashop_product_product and product_product
+    always_available = fields.Boolean(
+        string='Active',
+        default=True,
+        help='if check, this object is always available')
+    quantity = fields.Float(
+        string='Computed Quantity',
+        help="Last computed quantity to send on PrestaShop."
```
elicoidal: s/on/to

connector_prestashop/models/product.py
```diff
@@ -0,0 +1,194 @@
...
+        required=True,
+        ondelete='cascade',
+        string='Template',
+    )
+    # TODO FIXME what name give to field present in
+    # prestashop_product_product and product_product
+    always_available = fields.Boolean(
+        string='Active',
+        default=True,
+        help='if check, this object is always available')
```
elicoidal: If checked the product is considered always available,

connector_prestashop/models/product.py
```diff
@@ -0,0 +1,194 @@
...
+        string='Description',
+        translate=True,
+        help="Description html from PrestaShop",
+    )
+    description_short_html = fields.Html(
+        string='Short Description',
+        translate=True,
+    )
+    date_add = fields.Datetime(
+        string='Created At (on PrestaShop)',
```
elicoidal: s/on/in (several)

connector_prestashop/models/product.py
```diff
@@ -0,0 +1,194 @@
...
+        comodel_name='prestashop.shop',
+        string='Default shop',
+        required=True
+    )
+    link_rewrite = fields.Char(
+        string='Friendly URL',
+        translate=True,
+    )
+    available_for_order = fields.Boolean(
+        string='Available For Order',
```
elicoidal: Available for Order Taking

connector_prestashop/models/product.py
```diff
@@ -0,0 +1,194 @@
...
+        default=True,
+    )
+    show_price = fields.Boolean(string='Show Price', default=True)
```
elicoidal: s/show/display

connector_prestashop/models/product.py
```diff
@@ -0,0 +1,194 @@
...
+class PrestashopGroupsPricelist(models.Model):
+    _name = 'prestashop.groups.pricelist'
+    _inherit = 'prestashop.binding'
+    _inherits = {'product.pricelist': 'openerp_id'}
+
+    openerp_id = fields.Many2one(
+        comodel_name='product.pricelist',
+        required=True,
+        ondelete='cascade',
+        string='Openerp Pricelist',
```
elicoidal: Odoo (several)

guewen:
Thanks! Guess I know what I'll do today :) I'll do the review while working on the migration to 9.0

oca_dependencies.txt
```diff
@@ -0,0 +1 @@
+connector
```
guewen, line 1: the dependencies should be moved from `.travis.yml` to here

guewen:
I think the `requirements.txt` file at the root has not the correct version of prestapyt

connector_prestashop/product.py
```diff
@@ -306,34 +342,65 @@ def procure_method(self, record):
...
 
 @prestashop
+class ProductCategoryAdapter(GenericAdapter):
+    _model_name = 'prestashop.product.category'
+    _prestashop_model = 'categories'
+    _export_node_name = 'category'
+
+
+@prestashop
 class ProductInventoryExport(ExportSynchronizer):
```
guewen: should be `Exporter`

connector_prestashop/stock_tracking.py
```diff
@@ -34,6 +13,7 @@
 
 _logger = logging.getLogger(__name__)
 
+
 @prestashop
 class PrestashopTrackingExport(ExportSynchronizer):
```
guewen: should be `Exporter`

guewen:
A grep on `ExportSynchronizer` gives a few occurrences to replace with `Exporter`:

```
connector_prestashop/product.py
12:from openerp.addons.connector.unit.synchronizer import ExportSynchronizer
381:class ProductInventoryExport(ExportSynchronizer):

connector_prestashop/stock_tracking.py
8:from openerp.addons.connector.unit.synchronizer import ExportSynchronizer
18:class PrestashopTrackingExport(ExportSynchronizer):

connector_prestashop/sale.py
8:from openerp.addons.connector.unit.synchronizer import (ExportSynchronizer)
83:class SaleStateExport(ExportSynchronizer):
```

connector_prestashop/unit/backend_adapter.py
```diff
@@ -31,6 +8,12 @@
 from ..backend import prestashop
 
 _logger = logging.getLogger(__name__)
+handler = logging.FileHandler('/opt/odoo/v8/adapter_log.log')
+handler.setLevel(logging.INFO)
+formatter = logging.Formatter(
+    '%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+handler.setFormatter(formatter)
+_logger.addHandler(handler)
```
guewen: I guess this can be removed :)

connector_prestashop/models/delivery.py
```diff
@@ -0,0 +1,60 @@
...
+    prestashop_bind_ids = fields.One2many(
+        comodel_name='prestashop.delivery.carrier',
+        inverse_name='openerp_id',
+        string='PrestaShop Bindings',
+    )
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        string='Company',
+        index=True,
+        required=True,
```
guewen: Is it really necessary to have a required company?

connector_prestashop/backend.py
```diff
@@ -0,0 +1,13 @@
...
+import openerp.addons.connector.backend as backend
+
+
+prestashop = backend.Backend('prestashop')
+# version < 1.6.0.9
+prestashop1500 = backend.Backend(parent=prestashop, version='1.5')
+# version 1.6.0.9 - 1.6.0.10
+prestashop1609 = backend.Backend(parent=prestashop, version='1.6.0.9')
+# version >= 1.6.0.11
+prestashop16011 = backend.Backend(parent=prestashop, version='1.6.0.11')
```
guewen, line 13: Is there a difference in the PrestaShop's API between 1.6.0.9 and 1.6.0.11 that justify to have different backends?

pedrobaeza: Yeah, it was a nightmare. They renamed a lot of PS objects in that versions.

guewen: Ouch! Good to know they don't use semantic versioning.

connector_prestashop/delivery.py
pedrobaeza: As the previous work was too much mixed in the authors thing, I have preferred to let all files without copyright for now instead of putting all the possible authors in all the files. The manifest gathers all the authors I have found.

connector_prestashop/models/delivery.py
pedrobaeza: Checking twice, I'll leave it non required for now.

connector_prestashop/models/partner.py
pedrobaeza: There are too much dependencies with this old name, so for now we have to live with it.

elicoidal: I understand: it is just that the longer we wait the more the dependencies.
Non-blocking anyway

connector_prestashop/unit/mapper.py
```diff
@@ -496,33 +446,28 @@ def backend_id(self, record):
...
 
     @mapping
     def total_tax_amount(self, record):
-        tax = float(record['total_paid_tax_incl'])\
-                - float(record['total_paid_tax_excl'])
+        tax = (float(record['total_paid_tax_incl']) -
+               float(record['total_paid_tax_excl']))
         return {'total_amount_tax': tax}
 
     def _after_mapping(self, result):
```
guewen: `after_mapping` is never called (the method called at the end of the mapping in `Mapper` is `finalize`)

connector_prestashop/models/partner.py
pedrobaeza: @guewen What do you think about renaming now to `odoo_id`? Will it interfere with your refactoring?

guewen: @pedrobaeza no problem on my side

connector_prestashop/models/mail_message.py
```diff
@@ -0,0 +1,27 @@
...
+        comodel_name='prestashop.mail.message',
+        inverse_name='openerp_id',
+        string='PrestaShop Bindings',
+    )
+
+
+class PrestashopMailMessage(models.Model):
+    _name = "prestashop.mail.message"
+    _inherit = "prestashop.binding"
+    _inherits = {'mail.message': 'openerp_id'}
```
pedrobaeza: @guewen, I'm thinking about removing this inherit by delegation, because it doesn't add anything relevant, as we have already all the information via openerp_id.field. What do you think?

guewen:
In `SaleOrderLineRecordImport`, the few following lines should be updated to the new API:

``` python
         self.mapper.convert(self.prestashop_record)
         record = self.mapper.data
         record['order_id'] = order_id
```

connector_prestashop/models/mail_message.py
guewen: This `_inherits` is mainly used to write on the `mail.message` through the `prestashop.mail.message` in the importer. Do you have an idea to change that especially for `mail.message` or globally?

pedrobaeza: My idea was to change it globally, but I didn't think about the chatter question (anyway, it's not used currently as there's no track defined).

guewen: Using `_inherits` by delegation is by-design in the connector, so you have only 1 mapper/importer for both the `binding` object and the bound object ("normal" record), I'm not sure to see how you want to change that.

pedrobaeza: OK, I'm leaving as is for now. I still don't know enough the connector architecture (my colleague was the one in charge).

pedrobaeza:
> In `SaleOrderLineRecordImport`, the few following lines should be updated to the new API:

@guewen, can you tell me how that lines are translated to new API?

guewen:
@pedrobaeza Sure:

You get a `MapRecord` object with:

``` python
map_record = self.mapper.map_record(self.prestashop_record)
```

The `MapRecord` contains the source values and a reference to the mapper which will be used to transform the values when we ask for the values with:

``` python
record = map_record.values()
```

Or as we are creating the line, it's even better to gen the values with:

``` python
record = map_record.values(for_create=True)
```

As I'm on the topic, here is an underrated feature: you can pass any parameter in `.values()` which will then be accessible in the Mapper:

``` python
record = map_record.values(for_create=True, custom_value=1)

# [...]

    @mapping
    def value(self, record):
        if self.options.custom_value == 1:
            # do stuff
        else:
            # do other stuff
```

connector_prestashop/unit/import_synchronizer.py
```diff
@@ -80,38 +55,37 @@ def _validate_data(self, data):
...
-                      self.model._name, erp_id, self.prestashop_id)
-        return erp_id
+    def _create(self, data):
+        """ Create the OpenERP record """
+        # special check on data before import
+        self._validate_data(data)
+        model = self.model.with_context(connector_no_export=True)
+        model = str(model).split('()')[0]
+        binding = self.env[model].with_context(
+            connector_no_export=True).create(data)
```
guewen: This is really weird :)

The 4 lines above can be reduced to:

``` python
binding = self.env[model].with_context(
    connector_no_export=True
).create(data)
```

connector_prestashop/models/account.py
```diff
@@ -0,0 +1,76 @@
...
+            for invoice_line in invoice.invoice_line:
+                if invoice_line.product_id.id != discount_product_id:
+                    continue
+                amount = invoice_line.price_subtotal
+                partner = invoice.partner_id.commercial_partner_id
+                refund = self._find_refund(-1 * amount, partner)
+                if refund:
+                    invoice_line.unlink()
+                    line_replacement[invoice] = refund
+                    invoice.button_reset_taxes()
```
guewen: there seems to be no method called `button_reset_taxes`

connector_prestashop/account.py
```diff
@@ -102,14 +95,17 @@ def comment(self, record):
...
     def invoice_lines(self, record):
-        slip_details = record.get('associations', {}).get('order_slip_details', []).get('order_slip_detail', [])
+        slip_details = record.get(
+            'associations', {}
+        ).get('order_slip_details', []).get(
+            self.backend_record.get_version_ps_key('order_slip_detail'), [])
         if isinstance(slip_details, dict):
             slip_details = [slip_details]
         lines = []
         fpos_id = self.from_sale_order(record)['fiscal_position']
```
guewen: does not concern this line exactly but the `from_sale_order` method: it reads `fiscal_position` from sale.order which has been renamed to `fiscal_position_id'`. Anyway, it's useless to return a dict from `from_sale_order` instead of the recordset.

connector_prestashop/account.py
```diff
@@ -154,20 +149,21 @@ def _invoice_line_shipping(self, record, fpos):
...
+        binder = self.binder_for('prestashop.sale.order')
+        sale_order_id = binder.to_odoo(record['id_order'])
+        sale_order = self.env['prestashop.sale.order'].browse(sale_order_id)
 
         if not sale_order.carrier_id:
             return None
 
-        sale_order_line_ids = self.session.search('sale.order.line', [
+        sale_order_line_ids = self.env['sale.order.line'].search([
             ('order_id', '=', sale_order_id),
```
guewen: `sale_order_id` here is the id of the binding (`prestashop.sale.order`), not the "unwrapped" id, should be `sale_order.openerp_id.id`

connector_prestashop/account.py
```diff
@@ -220,50 +216,40 @@ def _invoice_line(self, record, fpos):
...
-            self.session.cr,
-            self.session.uid,
-            partner_id,
-            context=context
-        )
+        binder = self.binder_for('prestashop.res.partner')
+        partner_id = binder.to_odoo(record['id_customer'])
+        partner = self.env['prestashop.res.partner'].with_context(
+            company_id=self.backend_record.company_id.id).browse(partner_id)
         return {'account_id': partner.property_account_receivable.id}
```
guewen: renamed to `property_account_receivable_id`

connector_prestashop/account.py
```diff
@@ -220,50 +216,40 @@ def _invoice_line(self, record, fpos):
         }
```
guewen: in `_invoice_line`, `property_account_income_categ` has been renamed to `property_account_income_categ_id`

connector_prestashop/account.py
guewen: Or better, directly get the unwrapped record with:

``` python
        sale_order_id = binder.to_openerp(record['id_order'], unwrap=True)
```

connector_prestashop/product_combination.py
```diff
@@ -0,0 +1,402 @@
...
+                result[attribute] = main_template[attribute]
+        return result
+
+    def main_template(self, record):
+        if hasattr(self, '_main_template'):
+            return self._main_template
+        template_id = self.get_main_template_id(record)
+        self._main_template = self.env['prestashop.product.template'].browse(
+            template_id)
+        return self._main_template
```
guewen: this is a dangerous thing, the same mapper can be used for several `MapRecord`, nothing should be stored on `self`. (not sure to see the benefits BTW)

connector_prestashop/product.py
```diff
@@ -382,38 +456,35 @@ def _get_quantity(self, record):
...
+    def _get_template(self, record):
         if record['id_product_attribute'] == '0':
-            binder = self.get_binder_for_model('prestashop.product.product')
-            return binder.to_openerp(record['id_product'], unwrap=True)
-        binder = self.get_binder_for_model('prestashop.product.combination')
-        return binder.to_openerp(record['id_product_attribute'], unwrap=True)
+            binder = self.binder_for('prestashop.product.template')
+            return binder.to_odoo(record['id_product'], unwrap=True)
+        binder = self.binder_for('prestashop.product.combination')
+        return binder.to_odoo(record['id_product_attribute'], unwrap=True)
```
guewen: the method returns in one case a `product.template` and in the other case a `product.product`, it will be an issue for the wizard `stock.change.product.qty` called on a wrong id

connector_prestashop/unit/import_synchronizer.py
pedrobaeza: Well, I have put:

``` python
binding = self.model.with_context(
    connector_no_export=True
).create(data)
```

connector_prestashop/models/account.py
pedrobaeza: In v8 it is: https://github.com/odoo/odoo/blob/75a196972e9e1056c8ab9c299ddfc8e2eaafd812/addons/account/account_invoice.py#L637

connector_prestashop/account.py
pedrobaeza: The field name is correct: https://github.com/odoo/odoo/blob/75a196972e9e1056c8ab9c299ddfc8e2eaafd812/addons/sale/sale.py#L248

Are you seeing to 9.0 version?

connector_prestashop/account.py
pedrobaeza: Nop: https://github.com/odoo/odoo/blob/75a196972e9e1056c8ab9c299ddfc8e2eaafd812/addons/account/partner.py#L342

connector_prestashop/account.py
pedrobaeza: That's again in 9.0, not in 8.0

connector_prestashop/product.py
pedrobaeza: In both cases it returns `product.template`, but PS structure changes when you have combinations, so we need to change the field to query in that case.

pedrobaeza:
Comments attended and all green.

connector_prestashop/models/account.py
guewen: Of course, I forgot that I was on the 9.0 one :) sorry for the confusion.

connector_prestashop/account.py
guewen: You are right, forgot what I said!

guewen: But I still think this can be simplified by just directly using the order from `_get_order` 

``` python
fpos = self._get_order(record).fiscal_position
```

guewen:
@pedrobaeza sorry for the confusion between 8.0 and 9.0, I'll be cautious. Thanks for the fixes.

connector_prestashop/product.py
guewen: No, the binder for `prestashop.product.combination` returns a `product.product`:

``` python
class PrestashopProductCombination(models.Model):
    _name = 'prestashop.product.combination'
    _inherit = 'prestashop.binding'
    _inherits = {'product.product': 'openerp_id'}

    openerp_id = fields.Many2one(
        comodel_name='product.product',
        string='Product',
        required=True,
        ondelete='cascade',
    )
```

connector_prestashop/unit/binder.py
```diff
@@ -0,0 +1,115 @@
...
+        'prestashop.address',
+        'prestashop.res.partner.category',
+        'prestashop.res.lang',
+        'prestashop.res.country',
+        'prestashop.res.currency',
+        'prestashop.account.tax',
+        'prestashop.account.tax.group',
+        'prestashop.product.category',
+        'prestashop.product.image',
+        'prestashop.product.product',
```
guewen: `prestashop.product.product` to remove, this model does no longer exist

connector_prestashop/unit/mapper.py
```diff
@@ -697,31 +638,46 @@ def backend_id(self, record):
...
     def product_id(self, record):
         if record['id_product_attribute'] != '0':
-            binder = self.get_connector_unit_for_model(Binder, 'prestashop.product.combination')
-            return {'product_id': binder.to_openerp(record['id_product_attribute'], unwrap=True)}
-        binder = self.get_connector_unit_for_model(Binder, 'prestashop.product.product')
-        return {'product_id': binder.to_openerp(record['id_product'], unwrap=True)}
+            binder = self.unit_for(Binder, 'prestashop.product.combination')
+            return {'product_id': binder.to_odoo(
+                record['id_product_attribute'], unwrap=True)}
+        binder = self.unit_for(Binder, 'prestashop.product.product')
```
guewen: `prestashop.product.product` should be another model
`self.unit_for(Binder, ...)` can be replaced by `self.binder_for(...)`

sergio-teruel:
@pedrobaeza @guewen In line: https://github.com/OCA/connector/blob/15e0e45731971d68a4063e784f59b67025e95d70/connector/connector.py#L403 of connector module is defined 'to_openerp' and in binder.py there is one call (https://github.com/OCA/connector/blob/c3fb8261515187421c6ff37aa925984f46bbee36/connector/unit/mapper.py#L232-L232) so i think that we should be rename to_openerp to to_odoo and make a new method to_openerp calling to_odoo to avoid errors i other modules as connector_magento

pedrobaeza:
@sergio-teruel I did the renaming in https://github.com/OCA/connector-prestashop/pull/22/commits/fa7da5c43a7cf88004d8d713b41803e1affee15f

guewen:
@sergio-teruel  As it it a breaking change, the renaming of `to_openerp` to `to_odoo` should be done in the `connector` module in v10.

pedrobaeza:
cc @Tecnativa

connector_prestashop/unit/import_synchronizer.py
```diff
@@ -551,20 +469,37 @@ def _import_dependencies(self):
...
+                'order_id': erp_order.odoo_id.id,
+                'product_id':
+                    erp_order.odoo_id.carrier_id.product_id.id,
+                'price_unit':  shipping_total,
+                'is_delivery': True
+            })
+        erp_order.odoo_id.recompute()
+        return True
+
     def _check_refunds(self, id_customer, id_order):
```
guewen: this method is never called

pedrobaeza:
@guewen, @sergio-teruel is right. We can include both as an alias to keep compatibility. I renamed the method calls without seeing that the method was in connector, not here.

llacroix:
Not sure how to merge request is actually passing all checks... but this module can't be merged in 8.0 yet. The problem is that the connector module doesn't have a `to_odoo` method yet and this module is using it instead of `to_openerp`.

sergio-teruel:
@llacroix you are right.. tomorrow i will a pr in connector repository as 9.0 to solve this..
Cc @guewen @simahawk @pedrobaeza

llacroix:
There is also an other bug that I fixed. There should be a check:

```
diff --git a/connector_prestashop/models/res_country/importer.py b/connector_prestashop/models/res_country/importer.py
index a658c2b..8932f17 100644
--- a/connector_prestashop/models/res_country/importer.py
+++ b/connector_prestashop/models/res_country/importer.py
@@ -12,7 +12,12 @@ class CountryImporter(DirectBinder):
     _ps_field = 'iso_code'

     def _compare_function(self, ps_val, erp_val, ps_dict, erp_dict):
-        if len(erp_val) >= 2 and len(ps_val) >= 2 and \
-                erp_val[0:2].lower() == ps_val[0:2].lower():
+        if (
+            erp_val and
+            ps_val and
+            len(erp_val) >= 2 and
+            len(ps_val) >= 2 and
+            erp_val[0:2].lower() == ps_val[0:2].lower()
+        ):
             return True
         return False
```

I had an error where erp_val was False and in that case, python will raise an error because bool object do not have a `__len__` attribute.

sergio-teruel:
@llacroix please can you make a PR in repo https://github.com/Tecnativa/connector-prestashop ? and I will merge this fix..
Thanks

simahawk:
@sergio-teruel @llacroix the work is already there https://github.com/OCA/connector/commit/2a594891d861db7bd032fe1466bc06a1b3a780eb

you can backport it to v8 :)

llacroix:
There's also a problem with the connector but I'm not sure how to fix it yet. I believe the problem is caused by odoo creating multiple instances of the same prestashop mapping objects. The object for categories seems to have a uniq constraint which effectively prevent that from happening. 

My guess is that if you export multiple products with multiple job running at the same time, there is a chance that the connector will create multiple bindings for the same object. After that, any call to to_backend for those bindings will fail because it will fail the `ensure_one` predicate.

From my observation, I had the same problem with categories except the constraint made the job fail. Rerunning the job worked because it didn't try to create what already existed. Objects without constrains on `(backend_id, odoo_id)` will fail the `ensure_one` predicate in `to_backend` for the search of bindings.

I made a quick fix, by selecting only the first binding, yet adding a limit=1 to the search could also help...Yet that wouldn't really solve the concurrency issue. The only thing we can really do is add a constraint on the relation to prevent duplicate bindings. Job will fail but can be restarted and should work in theory. There could be a way to make it retry the job if a constraint error is raised, but I'm not sure how to do that.

llacroix:
@sergio-teruel I did the PR

simahawk:
@llacroix @sergio-teruel I think this has been fixed by @guewen here https://github.com/simahawk/connector-prestashop/blob/9.0-mig%2Btecnativa-imp/connector_prestashop/unit/importer.py#L186

IMO it would be better to backport our improvements (see https://github.com/guewen/connector-prestashop/pull/5) to v8 before going on with anything else.
Also, files structure in the module has changed to be improved, so if we want to port some fix/imp forward later on this will be more difficult if v8 and v9 WIP continue diverging.

llacroix:
@simahawk well obivously... Not sure about v9 and v10 yet, but if the modules can be backported, then it would be very nice. When you say that the file structure changed, do you mean it changed to help separate odoo version dependant code from the connector part. Technically most of the logic of the connector shouldn't be tied to a particular odoo version after v8. 

Hopefully, backporting won't break the system that are actually running.

simahawk:
@llacroix moving files or code is not because of odoo version, is just for the sake of having a well-organized module :)
I think that the current v9 version works on v8 too, but we are not using it on v8 so we cannot be sure about it unless some of you tests it :)

llacroix:
@simahawk does it support product export and stock export?

simahawk:
@llacroix this has been added by Tecnativa here https://github.com/OCA/connector-prestashop/pull/23 and has been merged and improved on v9 in the same PR I mentioned before https://github.com/guewen/connector-prestashop/pull/5

sergio-teruel:
@simahawk I did the same of this commit OCA/connector@2a59489 but @lmignon is right so i think that we must change it in v9 too.

llacroix:
@simahawk  I was asking because the PR23 doesn't completely work for me. I have some issue exporting stock. It seems to fail and the error is passing in the "strange" try/except block that ignore the errors. As far as I remember it was an error as if the url to the resource was pointing to something that doesn't exist.

I'm testing with prestashop 1.6.1.7

sergio-teruel:
@llacroix but connector_prestashop_catalog_manager was installed?

simahawk:
@llacroix sounds like an issue w/ endpoint names. See https://github.com/simahawk/connector-prestashop/blob/8a05bf86727736900f3678169523847f978b8bad/connector_prestashop/models/prestashop_backend/common.py#L224

Also, we are replacing silent errors with checkpoints, like here https://github.com/guewen/connector-prestashop/pull/5/commits/c8d7a8fcc66a1fc20ef8664228a0b3a0d2948727

llacroix:
@sergio-teruel yes, I had #22 and #23 merged. @simahawk Ah that might be the problem.

connector_prestashop/models/product_product/common.py
```diff
@@ -0,0 +1,241 @@
...
+    @api.multi
+    def update_prestashop_qty(self):
+        for product in self:
+            if product.product_variant_count > 1:
+                # Recompute qty in combination binding
+                for combination_binding in product.prestashop_bind_ids:
+                    combination_binding.recompute_prestashop_qty()
+            # Recompute qty in product template binding if any combination
+            # if modified
+            for prestashop_product in \
```
llacroix, line 98: This part of code should be removed. It may create concurrency errors with the prestashop_product_template being modified by multiple variants being created. The concurrency error will make the transaction fail and as a result odoo will loose the binding that has just been created previously. 

There are 2 possible results to this:
- 2 combinations are created for 1 variant (considering the second try of the job works)
- 1 combination created but second try doesn't work (constraint error)

Not sure yet why I get both kind of errors... in 1 case I get the jobs to end without failure but I have an extra combination on prestashop that isn't bound to anything... the other case makes the error fail completely... Removing the variant on prestashop will make the job continue...

Anyhow, I'm not sure we should update the quantity on the template since the quantity is actually not stored in one variant... So even if we update the quantity, it seems that one quantity override the other without actually computing the sum of all variants available... which is kind of erroneous.

llacroix:
@simahawk Finally solved the problem, It was a configuration error, apparently a module (mod_security) was blocking PUT request on apache... 

That said, I found the problem with concurrency. While exporting products using PR23, it updates the product quantity, but this PR update template quantities at the same times. 

Finally when exporting multiple variants, it end up updating the template quantity at the same time and then a concurrency error is raised.

llacroix:
Also regarding the export of quantities, I'm pretty sure the quantities exported are wrong. Since we can only create one prestashop connector by warehouse, the quantity exported by products should be warehouse dependant. Now, it seems to export the quantity available overall. This might work since we're not exporting stock by warehouse but that kind of make the warehouse binding by backend useless.

pedrobaeza:
@llacroix, you select which locations to be exported to PrestaShop, so this not might be a problem.

llacroix:
@pedrobaeza  Not sure it does. 
https://github.com/OCA/connector-prestashop/pull/22/files#diff-2dc84c07ca9ac7470689ffd6c16c3402R162 
unlike here: 
https://github.com/OCA/connector-prestashop/pull/22/files#diff-0898e50a87607fc0078e5184789faec2R115

pedrobaeza:
Uhm, then this version is not the latest one. @sergio-teruel where is the version with the stock by location selectable?

sergio-teruel:
hi @pedrobaeza @llacroix the lastest repository is https://github.com/simahawk/connector-prestashop/tree/9.0-mig+tecnativa-imp I do PR to @simahawk and he does PR to @guewen 
To select location to synchronize with PS only you need go to location and check **"Synchronize with PrestaShop"**.
Our lastest commits is here: https://github.com/Tecnativa/connector-prestashop/tree/9.0-PR-connector_prestashop-SM.
In a few days i will do a new PR with a lot improvements.

pedrobaeza:
But isn't this available for 8.0?

llacroix:
By the way, is there a real use to the config warehouse_id in the prestashop backend. I just tested that and to my surprise, it seems to not do what I was expecting..

I believed that I could create a backend by warehouse in the system. Each backend would create its own warehouse and then it would be possible to manage quantities of stock by warehouse on prestashop... Unfortunately here's how it works. If I do that, it will reexport all products by warehouse in prestashop because product binding are based on [backend/odoo_id] which means that even though backends might point to the same backend, odoo will create multiple identical products in the same warehouse in prestashop but with quantities of odoo by warehouse. 

It's not really good when you have multiple warehouse in our case, we'd end up with 5 times the same product... but technically this bug might not be a big issue in our usecase.. All I see is the warehouse_id config in the backend makes no sense since you can't export quantities of particular location in different warehouses and you can't really have multiple backends anyway.

Also the current code I posted above which count stock seems to be limited to the "stock location" which might be much more complex in some cases.

pedrobaeza:
Well, we don't have such cases of multiple warehouses. Feel free to improve the current code to accommodate the connector to your use case.

llacroix:
@sergio-teruel  looking at your code in v9. I'm not sure it works for product.product either like in v8. Anyway, wouldn't it be better to simply backport v9 to v8. I'm not certain of the changes made to v9 but it would be good if most of the work was "similar" so bugs can get fixed either way. 

I have more fixes btw, can I make a PR on your branch again?

pedrobaeza:
If you can try the version 9, then we can join efforts to refine that version, and the we can backport it to v8 easily.

guewen:
@sergio-teruel @pedrobaeza  Do you know where we can find the valid name of the resources returned by `get_version_ps_key`? I tried to find in PrestaShop source code without luck. I want to check because I had to change again a key and I don't want to change but introduce a regression for another version. Plus, one of the backend version (1.6.1.12) doesn't even exist (shouldn't it be 1.6.0.12? or 1.6.1.0?)

edit: I see it is a different version in the model than in backend.py, it is 1.6.1.2.

sergio-teruel:
@guewen in 1.6.0.9 the name are plural

> [-] WS : #PSCFV-10283, #PSCFV-7498 Fix product images behavior [-] WS : #PSCFV-11895, #PSCFV-10748 Fix ampersand issue [-] WS : Fix node_type behavior for associations, now the node_type attribute is linked to plural entities name (eg: category is now categories)

in 1.6.0.11 the names are singular:

> [-] WS : Fix nodeType attribute, now singular [*] WS : Add "api" attribute, plural to name entity

but ... mhuuu. perhaps in 1.6.1.2 are in singular too, i don't find where I read this... :confused: 
Perhaps we can use the api attribute?? bat in older versions is not present.

One question , I need connect with another url (multi-store shop) i tried do that here [https://github.com/Tecnativa/connector-prestashop/blob/9.0-FULL-migration-connector-prestashop/connector_prestashop/unit/backend_adapter.py#L99](url) but i have seen that you remove this in youu branch, I updated the session context with kwards values becouse i don't know how to propagate the kawards at this point... Can you tell me some way to do it?

guewen:
@sergio-teruel 
For the version 1.6.0.9, I saw that too and I'm sure of it. That's for the 1.6.1.x that I'm really not sure.

For the `shop_url`, I was wondering, isn't a `shop_url` related to a `prestashop.shop`? In that case, shouldn't you link the binding to shop?

Anyway, for passing custom attributes, you can either:

If you have to pass it in some specific places, inherit the ConnectorUnit and get the argument from the `kwargs`, then you can do whatever you want with it (pass to backend adapter or another unit, set in the context, ...):

```python
@custom
class CustomSaleOrderImporter(SaleOrderImporter):
    _model_name = 'prestashop.sale.order'

    def __init__(self, environment):
        super(CustomSaleOrderImporter, self).__init__(environment)
        self.shop_url = None

    def _get_prestashop_data(self):
        """ Return the raw prestashop data for ``self.prestashop_id`` """
        assert self.shop_url
        return self.backend_adapter.read(
            self.prestashop_id, shop_url=self.shop_url
        )

    def _import(self, binding, **kwargs):
        self.shop_url = kwargs.pop('shop_url', None)
        assert self.shop_url, "must receive a shop_url"
        # or set a key in the context or whatever
        return super(CustomSaleOrderImporter, self)._import(binding, **kwargs)

```

The other option is to set it in a custom `ConnectorEnvironment`. That's maybe what you want because it is then shared horizontally:

```python

class PrestashopBackend(models.Model):
    _inherit = 'prestashop.backend'

    @api.multi
    def get_environment(self, model_name, session=None, shop_url=None):
        self.ensure_one()
        if not session:
            session = ConnectorSession.from_env(self.env)
        # you could take shop_url from the context as well
        return CustomConnectorEnvironment(self, session, model_name, shop_url=shop_url)

class CustomConnectorEnvironment(ConnectorEnvironment):

    _propagate_kwargs = ['shop_url']

    def __init__(self, backend_record, session, model_name,
                 shop_url=None):
        super(CustomConnectorEnvironment, self).__init__(
            backend_record, session, model_name
        )
        self.shop_url = shop_url  # is available in any ConnectorUnit in self.connector_env.shop_url
```

sergio-teruel:
Hi @guewen thanks for your help the second way is perfect for me but in function import not pass kwargs

```python
@job(default_channel='root.prestashop')
def import_record(
        session, model_name, backend_id, prestashop_id, **kwargs):
    """ Import a record from PrestaShop """
    backend = session.env['prestashop.backend'].browse(backend_id)
    env = backend.get_environment(model_name, session=session)
    importer = env.get_connector_unit(PrestashopImporter)
    importer.run(prestashop_id, **kwargs)
```
In line '**env = backend.get_environment(model_name, session=session)**' the kwargs are not been passed. So my custom enviroment not is inicializated correctly. 
Would we pass kwargs here??

guewen:
@sergio-teruel maybe we could an argument `connector_env_options` which is a dict and is passed as kwargs to the connector environment?

pedrobaeza:
After a long testing period, I'm going to merge this one as is very stable and possible patches can come as other PRs.

witoszek:
Hello 

II've got an error and i don't know how i can fix it :
if i Click on "Synchronize Base Data" button  appear : 
An ERP record with same ID already exists on PrestaShop.
Ive not added Products in prestashop-
Version of prestashop 1.705.
Ive got product with demo installation,
Ive re do it an api webkey and nothing and re install prestashop and nothing.

PR #23, opened 2016-09-05
[8.0][ADD] connector_prestashop_catalog_manager

pedrobaeza:
# 
# Catalog Manager for Odoo PrestaShop Connector

This module is an extension for _connector_prestashop_. With it, you will be
able to manage your catalog directly from Odoo:
- Create/modify attributtes and values in Odoo and push then in PrestaShop.
- Create/modify products and push them in PrestaShop.
- Create/modify products variants and push them in PrestaShop (combinations).
- Create/modify category and push them in PrestaShop.
- Create/modify image and push then in PrestaShop.
# Known issues / Roadmap
- Tests.


pedrobaeza:
WIP until #22 is merged.

Relevant commit is https://github.com/OCA/connector-prestashop/pull/23/commits/823921952416e8a641b289a1f124b92b90197611

pedrobaeza:
cc @Tecnativa

simahawk:
@pedrobaeza looks like this module push product attrs as features instead of combination: is that correct? If yes, is there any plan to move to combination?

pedrobaeza:
No, this is working with standard PS combinations. Why do you say that?

simahawk:
@pedrobaeza sounds like we were looking in the wrong place... found the right code, sorry for the noise :S 
I will test really the whole thing in next days.

pedrobaeza:
And I'm merging also this one with the same logic. Now let's go for the version 9!

PR #24, opened 2016-09-13
[WIP][9.0] Migrate prestashop_connector

guewen:
Notes:
- I use https://github.com/guewen/odoo-prestashop-dev to setup the project with
  Prestashop + Odoo and the branches
- To run the tests, the additional eggs `vcrpy` and `freezegun` are required
  (already installed in the docker composition above)
- I use a development branch of Prestapyt that replaces `httplib2` by
  `requests` so we can record test fixtures with `vcrpy`. The branch is
  https://github.com/guewen/prestapyt/tree/use-requests and is already
  installed in the docker composition above

Still to do:
- [x] review the silent `try: except: pass` which seem questionable
- [x] use requests in prestapyt so we can use vrc.py for the tests (https://github.com/prestapyt/prestapyt/pull/37 should be merged and a new release of prestapyt published when we are confident with the changes)
- [ ] add tests (with vcr.py) (started)
- [x] investigate reason for the 'virtual' model `_import_stock_available` instead of using `prestashop.product.template` (no change)
- [ ] handle sale order states better according to odoo 9.0 changes: the 'done' state is meaningless for PrestaShop, but we'd want to know when an order is delivered and / or invoiced
- [x] port the last changes from the 8.0 version
- [x] import of refunds seems deeply broken
- [ ] every synchronization should be tested, reviewed and get a test suite
- [x] make a pass through the TODOs and FIXMEs
- [x] port the catalog manager module
- [ ] Add debug option for the API
- [ ] Add some matching options for taxes / product / partner
- [ ] kanban view for the backend (see woo-commerce)


flotho:
Hi @guewen , @pedrobaeza 

We're working with @gaelTorrecillas on @guewen  PR and have some functionnal questions to submit to you.
Our point is that we have multiple backends to synchronize to and that causes some difficult points when for example you have the same tax / barcode that already exists etc...
We would like to propose a generic method that will match if an object already exists regarding criteria that will be implemented for each specific object.
If the method match a record we would like to propose to bind this PS object to the existing Odoo Object for this specific backend. 
Do you approve this, do you consider it as useful, do you recommend to add an option for activating this behaviour?
Regards

guewen:
@flotho I think that the matching method will change from company to company. I'm not sure we can have generic matching methods, at least not for all the models (see the discussion on https://github.com/OCA/connector-magento/pull/219 for example).

For the implementation, it should just be a matter of setting an existing record in the `openerp_id` in the Mapper: https://github.com/OCA/connector-magento/pull/219/files

flotho:
Another question regarding tax.rule.
From your point of view, would it be useful to use the fiscal position to deal with country dependencies? Or is it better to let the object "outside" the odoo core?
regards

pedrobaeza:
The problem is that how PS handles taxes and how Odoo do it is too different. You will need to create a lot of intermediate objects to map equally all, and this will only serves for managing master data in Odoo and for nothing more.

guewen:
I'm fine with the fiscal positions

guewen:
> I'm fine with the fiscal positions

@pedrobaeza knows better than me about the complexity of mapping them, so listen to him :)

On the Magento connector, fiscal positions need to be created manually on Odoo to match the ones of Magento. That's not something you change often, so that's not a big deal. And the order cannot be confirmed if the tax amounts do not match.

flotho:
@guewen , regarding the matching in general, thanks for the tip, very usefull.
We would like to implement this on Tax / ProductProduct and maybe other objects in the future. Do you recommend to add options in the PS backend configuration ?
Regards

pedrobaeza:
Options would be good for flexibility obviously, but at least include match/not match boolean.

flotho:
@guewen, we would like you to add some tasks (@gaelTorrecillas will handle this):
- Add debug option for the API
- Add some matching options for taxes / product / partner

flotho:
@gaelTorrecillas, could you also have a look at this task : "port the last changes from the 8.0 version"
The principle is that @pedrobaeza made a lot of job regarding the v8 migration with dedicated features for some PS versions. have a look there https://github.com/OCA/connector-prestashop/pull/22

pedrobaeza:
You can also do a kanban view like the one that our OCA partner is doing in connector-woocommerce, which is great.

flotho:
@pedrobaeza, could you point out the resource , some images and also some explantions of the purpose of the Kanban view?
regards

pedrobaeza:
Look to this tweet: https://twitter.com/parthivpatel/status/769211777313746949 and you can ask @parthivgls for more details.

parthivgls:
@flotho it's more about creating a dashboard interface for the store, so that you can which quick actions needed.

pedrobaeza:
@parthivgls, I think we can create the generic backends view for all the connectors (from the main model), and let each module to overwrite template view and add elements with `t-if` conditions. You can see an example in Odoo itself with accounting dashboard, which starts from account.journal model to present totally different kanban elements if the journal is cash or bank or if the journal is for sale/purchase.

flotho:
@parthivgls , thanks for the explanation, really useful
@pedrobaeza , for sure a generic one could be really helpful

flotho:
@guewen , 1 more task to add please : 
- Add state on the backend and don't interact with it if it's not in production. @gaelTorrecillas will handle this

flotho:
@guewen , another task to add please : 
- Deal with default account_journal

pedrobaeza:
@flotho, please check last @parthivgls's tweet where you have a tip of how to do it: https://twitter.com/parthivpatel/status/776426743418716160

gaelTorrecillas:
hello,

@guewen thanks you for issue #25, I have an other question, I use your prestapyt, but it says this version isn't compatible with our version of prestashop 1.6.0.9, what is your version of prestashop used to test prestapyt in your deposit (ref : https://github.com/guewen/prestapyt/tree/use-requests)

'''
/usr/local/lib/python2.7/dist-packages/prestapyt/prestapyt.py:183: UserWarning: This library may not be compatible with this version of PrestaShop (1.6.0.9). Please upgrade/downgrade this library

guewen:
@gaelTorrecillas this is only a warning, which does not claim that it is not compatible but _might not be_. It means nobody thoroughly tested (or reported) the library with this version of PrestaShop. If you did and assert that it is compatible, a PR on prestapyt to increase the max. compatible version is welcome :)

gaelTorrecillas:
yes, tanks you @guewen , 
in fact, it works.
I works on debug, but I have a problem, I look for the cause

flotho:
Hi @guewen , you precise that you're using the product_multi_image module https://github.com/guewen/connector-prestashop/blob/9.0-migration-connector-prestashop/connector_prestashop/__openerp__.py#L17 from OCA, yet it seems to be available only via a PR that hasn't been merged : https://github.com/OCA/product-attribute/pull/161
Do you work with this PR ?
Thanks

flotho:
Hi @guewen , I proposed the PR https://github.com/guewen/connector-prestashop/pull/2 for matching the partners.

Additionnaly is there any other code on the account_payment_mode ?

regards

flotho:
@guewen , @pedrobaeza ,
I made this PR https://github.com/guewen/connector-prestashop/pull/2 to implements the matching on partners. The principle is fairly simple, you choose to match on partner and the code is browsing the PS fields available for this backend . Once you have chosen the field, the connector will search for the value of this field in the reference info of the partner in Odoo. I've tested it on a db with avout 500 customers and it works perfectly either for the partner and the adress. Any code review/merge would be appreciated so that my contribution won't still useless

regards

flotho:
@guewen , regarding payment mode, is it normal that there is no prestashop bindings in the common.py file?

flotho:
Hi @guewen , @pedrobaeza ,
Any chance to be reviewed on this :
- https://github.com/guewen/connector-prestashop/pull/2
- https://github.com/guewen/connector-prestashop/pull/3

simahawk:
@sergio-teruel this commit should fix the issue on `order_slip` for different versions https://github.com/OCA/connector-prestashop/pull/24/commits/d4065c1d34483b359cbf29fadbf3437a6cc7eaec

sergio-teruel:
@guewen I have a lot improvements as this fix
https://github.com/Tecnativa/connector-prestashop/blob/9.0-PR-connector_prestashop-SM/connector_prestashop/unit/binder.py#L42

And other modules...
Do I make a PR to @simahawk ??

guewen:
Thanks @sergio-teruel , I cherry-picked the last changes

guewen:
have you any idea on how to handle the order status with Odoo 9.0? ("handle sale order states better according to odoo 9.0 changes: the 'done' state is meaningless for PrestaShop, but we'd want to know when an order is delivered and / or invoiced")

@sergio-teruel @pedrobaeza @simahawk

PR #34, opened 2016-12-12
[9.0] Migrate prestashop_connector

guewen:
Follow-up of #24. I closed #24 because it was polluted with a lot of outdated comments. This is a fresh start after weeks of work from several contributors.

The connector now contains tests (maybe not the most exhaustive ones, but the first tests are the most valuable as they open the room for more tests when we'll discover bugs).

I think we should now merge this PR and continue with the possible improvements with new PRs, I mean, the migration work is done and more than that, it already contains improvements which are outside of the scope of a mere migration.

Before merging, a rebase (at least the remove the "!fixup" should be done, but please do not squash all the commits into one, because some commits contains explanations for the commit's changes.


simahawk:
@sergio-teruel @pedrobaeza hola :)
Do you use and sync `product_brand`? We need to sync the brand too (`manufacturer` on PS) and we are using [product_brand](https://github.com/OCA/product-attribute/tree/9.0/product_brand)

pedrobaeza:
We have made for v8 a layer for product_manufacturer, not for product_brand: https://github.com/Tecnativa/connector-prestashop/tree/8.0-PR-connector_prestashop_manufacturers/connector_prestashop_manufacturer

simahawk:
@pedrobaeza thanks, good to know! :)
I have another question: impact price. Sounds like [is calculated on odoo side](https://github.com/guewen/connector-prestashop/blob/9.0-migration-connector-prestashop/connector_prestashop/models/product_product/importer.py#L254) instead of being imported from PS with [its api](http://stackoverflow.com/questions/35580449/prestashop-special-price-for-product).
Is there an extra module or something you are working on that takes care of this? Otherwise we are going to implement it.

pedrobaeza:
In theory that field is handled by the connector base, but @sergio-teruel should say it.

sergio-teruel:
Hi Simone... Yes I added a new field in product.product, this field is
written in importer and you can manage for export too.

Impact_price field:
https://github.com/Tecnativa/connector-prestashop/blob/8.0-
connector_prestashop/connector_prestashop/models/
product_product/common.py#L26

​the lst_price is calculated here:
https://github.com/Tecnativa/connector-prestashop/blob/8.0-connector_prestashop/connector_prestashop/models/product_product/common.py#L31
​

In product_priduct mapper is stored:
https://github.com/Tecnativa/connector-prestashop/blob/8.0-
connector_prestashop/connector_prestashop/models/
product_product/importer.py#L249

And in catalog_manager is exported:
https://github.com/Tecnativa/connector-prestashop/blob/8.0-connector_prestashop_catalog_manager/connector_prestashop_catalog_manager/product_combination.py#L212
​

​I hope help you!! 😀😀​


Un saludo.

Sergio Teruel Albert.
Tecnativa

simahawk:
@sergio-teruel I see. Indeed this is what I found and what I named "calculated on odoo side" ;)
But on PS the impact price is not the price, it can a be a discount rule and comes in a special set of attributes for prices http://stackoverflow.com/questions/35580449/prestashop-special-price-for-product

sergio-teruel:
@simahawk Yes i know "specific_prices" PS model for set a price, but I have not worked with this... sorry..
but it could be integrated.

simahawk:
@sergio-teruel no need to feel "sorry" ;)
Next question: I don't get why the product type is set to "service" when has variants... [the explanation is not that clear](https://github.com/guewen/connector-prestashop/blob/9.0-migration-connector-prestashop/connector_prestashop/models/product_template/importer.py#L261). Do you know why?
To me it makes no sense, but maybe I'm missing something....

simahawk:
@sergio-teruel nope, sorry, it makes sense since you cannot have PS virtual prod w/ variants. 
The only issue is that in this way the product is always marked as "Stockable", so if you want the product to be "Consumable" you have to hack it. Probably a configuration in the backend will help...

simahawk:
@guewen tests are failing here because the sale order cannot be imported. The reason is: [the order has not been paid since... ](https://github.com/guewen/connector-prestashop/blob/d94f298a692eef0c7517e990e651a6962955ab7c/connector_prestashop/models/sale_order/importer.py#L116) bla bla and then [it raises this](https://github.com/guewen/connector-prestashop/blob/d94f298a692eef0c7517e990e651a6962955ab7c/connector_prestashop/models/sale_order/importer.py#L337) and the test fails.

Can't we use checkpoint to track this issue instead of breaking the import?

Plus, if the answer is "yes", let's consider that [the base importer handles "skip case" like this](https://github.com/guewen/connector-prestashop/blob/d94f298a692eef0c7517e990e651a6962955ab7c/connector_prestashop/unit/importer.py#L248) returning the skip message. This one is lost since nobody takes care of what `importer.run` returns.

Can't we simply add a checkpoint whenever we get back a skip message?

/cc @sergio-teruel @pedrobaeza

guewen:
@simahawk when `NothingToDoJob` is raised in a job, that's caught by the job runner and the job is automatically set to done with a message. The job should not fail. Maybe the test just has to take this in consideration. Nevertheless, I would not be against handling that like the skip message, because I was thinking about deprecating `NothingToDoJob` which is too error-prone because it cascades through a chain of importers.

Regarding the "skip message" which is lost: when a job function returns something, it's stored as a message in the job. So normally this message should not be lost (I didn't checked in the code here). We should have something like:

```python

@job
def import_record(blabla):
    # code
    return importer.run(blabla)
```

guewen:
@simahawk see https://github.com/OCA/connector/blob/9.0/connector/queue/job.py#L479

I checked and the jobs return nothing, so there is indeed never any skip message (or any other message) returned. This could be fixed by simply adding a `return` in the job functions.

simahawk:
@guewen yep, they don't return anything in PS case. I'll fix that. Anyway, you can see these messages only by digging into job queue failures and the average user (maybe the sale manager) won't know why the import of a sale order has failed and have to ask to the tech guy (you, me, etc ;)) to check.

pedrobaeza:
Guys, when do you think we can merge this and continue evolving with other PRs instead of acumulating in this one?

pedrobaeza:
The commit history frightens me a bit also...

simahawk:
@pedrobaeza I need this too ;)
The only blocker now: is the test that is failing. I'm working on this, then we can merge once for all.

pedrobaeza:
Do you accept me squashing commits by logical set before merging? Or do you prefer to do it yourselves? Or should I merge as is (some people has been upset recently by my "squashing obsession")?

simahawk:
@pedrobaeza I'll do it. Don't worry ;)

simahawk:
@pedrobaeza test is fixed. About squash: can I change my mind? :stuck_out_tongue_closed_eyes: 
I was trying to keep only really important fixes and '[add]' commits but I'm getting into a never-ending conflicts series. Can you kindly help with it? :heart:

rafaelbn:
Dear all, I would like to propose the optimistic merge of this PR. Reasons:
- We can add to "Known issues" section this to be care of
- As PR this module doesn't appear in Odoo apps or the list of modules in OCA this make that lots of people are buying different modules in the apps store and then we are missing that more people use this module and could give feedback, improve it or even migrate to new versions...

Please consider.

Thanks
Regards,
Rafael

rafaelbn (approved):
Functionally talking we have been using this module in production with customer during 4 months, so in my opinion 👍

simahawk:
@rafaelbn Happy to know you are already using it since months! 
Our schedule for going live w/ this was a bit delayed by our customer so we delayed merging this to see it working in real life. We fixed a couple of bugs and right now sounds strong enough. 

Finally, I agree w/ you: let's merge it! @guewen do you agree?

pedrobaeza:
We talked about merging some weeks ago, and I was about to merge this afternoon...

simahawk:
@pedrobaeza alright, go on :)

pedrobaeza:
All the commits are mixed with merge commits, and rebasing messes up everything with multiple conflicts, so I'm going to merge as is, although I would prefer to squash the commit story, but it's too much work in this case and this is not a perfect world.

Thanks to the C2C guys (@guewen & @simahawk) and @sergio-teruel from @Tecnativa for their tremendous work on this.

pedrobaeza:
I have also recreated 10.0 branch from 9.0 with the migration tool for having it for not having to put full commit history on migration.

PR #41, opened 2017-04-03
[add] filter of importable orders by state

simahawk:
If `backend_record.importable_order_state_ids` is valued
we check if current order is in the list.
If not, the job fails gracefully.

TODO: add tests.

### DIFF
--- a/connector_prestashop/models/sale_order/importer.py
+++ b/connector_prestashop/models/sale_order/importer.py
@@ -100,6 +100,7 @@ class SaleImportRule(ConnectorUnit):
                 "Process or create a new one.") % (ps_payment_method,
                                                    ps_payment_method))
         self._rule_global(record, payment_mode)
+        self._rule_state(record, payment_mode)
         self._rules[payment_mode.import_rule](self, record, payment_mode)
 
     def _rule_global(self, record, mode):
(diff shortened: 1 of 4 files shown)


simahawk:
Damn, this test is still failing randomly https://travis-ci.org/OCA/connector-prestashop/jobs/217990989#L1056.
@pedrobaeza @sergio-teruel @guewen any idea on how we can make this reliable?

simahawk:
@guewen after job restart the flappy test passed here too :smile:

guewen (approved):

connector_prestashop/models/sale_order/importer.py
```diff
@@ -117,6 +118,31 @@ def _rule_global(self, record, mode):
                                  'because it has not been paid since %d '
                                  'days' % (order_id, max_days))
 
+    def _rule_state(self, record, mode):
```
pedrobaeza, line 121: Why are you passing payment mode?

simahawk: just to respect the same signature. Not mandatory indeed, but if someone wants to customize it further based also on payment mode is feasible with it.

pedrobaeza: Well, it's not the normal to need that, but it doesn't hurt.

pedrobaeza (changes requested):
Please change module version number.

simahawk:
@pedrobaeza rebased and changed version to 1.0.4. Maybe better 1.1 since is a new feat?

pedrobaeza:
Well, technically speaking doesn't imply to update the module, but what you prefer. There's a Travis error now.

simahawk:
@pedrobaeza always the same stupid test on export qty which fails randomly... @guewen and I spent some time in trying to fix that but no luck :sob: 
I'll run the build again to see if it passes.

simahawk:
@pedrobaeza can we merge this? we are using it in prod since months...

pedrobaeza (approved):
OK, but we should fix this test ASAP. It's failing more and more and it distracts us from real errors

simahawk:
@pedrobaeza you are right and both me and Guewen spent time on that: no luck. Sounds like is not an easy one. Tnx for merging sir! :)

PR #43, opened 2017-04-04
7.0 corrections

bodi000:
a few improvements that block the functioning of 7.0


pedrobaeza:
Anyone more using this? It seems that some changes are due to PS version, but this version is very old and I don't know if this will be reviewed by proper colleagues.

PR #44, opened 2017-04-09
[8.0][FIX] connector_prestashop: Delete class manufacturer that provokes an error when import a product.

sergio-teruel:
[8.0][IMP] connector_prestashop: Set full_public_url in backend adapter.
cc @Tecantiva


pedrobaeza:
Why manufacturer extra make it crash? How do you handle it now?

sergio-teruel:
In v 9 we do not have any issue, so I have to investigate it, but until then, we have to remove that code.

pedrobaeza (approved):
OK then. Let's investigate it. As there's not more v8 maintainers of this, I merge it.

PR #45, opened 2017-04-09
[8.0][IMP] connector_prestashop_catalog_manager: change image url when export a new image or write one

sergio-teruel:
Can not be merged before this PR: https://github.com/OCA/connector-prestashop/pull/44
cc @Tecnativa


pedrobaeza (approved):
Merging as well, although I would like to protect imports next time.

PR #46, opened 2017-04-13
8.0 [FIX] connector prestashop minor fixes

sergio-teruel:
[8.0][FIX] connector_prestashop: Recompute qty in product template if product has not variants. This provoke extra writes.

[8.0][FIX] connector_prestashop_catalog_manager: Update always the default variant in las position. In recent ps versions set tow variants as default is forbidden

cc @Tecnativa


pedrobaeza (approved):
Fast-tracking it

PR #47, opened 2017-04-18
[9.0][IMP] account tax view

angelmoya:
On configuration process, when you set tax relation, on account.tax.group you can not set created taxes.

As a quick fix I only update view, to use many2many widget, but I think that it would be better to change this field from many2one to many2many field.

What do you think?

### DIFF
--- a/connector_prestashop/views/account_view.xml
+++ b/connector_prestashop/views/account_view.xml
@@ -31,7 +31,7 @@
         <field name="arch" type="xml">
             <form string="Account Tax Group">
                 <field name="name" colspan="4"/>
-                <field name="tax_ids" colspan="4"/>
+                <field name="tax_ids" colspan="4" widget="many2many"/>
                 <group string="PrestaShop Binding">
                     <field name="prestashop_bind_ids"
                            nolabel="1"/>


pedrobaeza:
@sergio-teruel please check

angelmoya:
@sergio-teruel @pedrobaeza Maybe I don't understand the scope of import account.tax.group from prestashop, because on spanish localization we have account.tax.group configured in one group, and if I set on imported tax groups then, as field is many2one, previous relation is deleted.

pedrobaeza:
I don't remember this exactly, but there's a reason I think about the divergences in tax handling between Odoo and PS. I let Sergio to specify more.

sergio-teruel:
@angelmoya In PS through webservice you can update a product tax with a rule group tax so the mapped record is tax.group, you can not update directly a product tax.. (I think...).
I think that is a bad idea change the tax_group field from Many2one to Many2many type, now I do not know the consequences of this change, perhaps @pedrobaeza can tell us if it is possible.. but mhuu i think that it is a hard change... :smile:

pedrobaeza:
Yes, I agree that the change is not going to work as is.

angelmoya:
OK @pedrobaeza  @sergio-teruel 

Now, with o2m on model and view, when you want to configure tax_groups imported from prestashop you are not available to use tax created by spanish localization, or you can set tax_group m2o field directly on tax?

With this improvement on view, you can set on tax groups the taxes that you want to use. Is right to change a tax from default taxes group to taxes group created by prestashop connector?

PR #50, opened 2017-05-09
[fix] allow user to read prestashop.product.category

simahawk:
**Current behavior**

* you are an inventory manager
* you go to "Product categories"
* you open a category
* KABOOM! No access to prestashop.product.category

**Desired behavior**

* no error pls!
* hide bindings if not connector manager group

NOTE: there are other views/models where we allow ppl to read records but maybe it would be better to hide connector bindings as I did here.


guewen (approved):

pedrobaeza:
cc @sergio-teruel

sergio-teruel (approved):

pedrobaeza (approved):

PR #53, opened 2017-05-18
Fix export when no location has flag prestashop_synchronized

guewen:
When there is no location with the flag prestashop_synchronized in the
location tree,

Current behavior:

The `location` key in the context is empty then the qty_available is
computed for *every* locations/warehouses.

Expected behavior:

If there is at least one location flagged, use it, otherwise, use all
the locations of the tree, starting from the selected location or
selected warehouse location.

When no location can be found (for instance because user selected a
location with a usage different than 'internal'), then it should just
fail and not return the stock of all warehouses.

Being able to synchronize without activating 'prestashop_synchronized'
is important: when used, this feature will trigger an export every time
a quant is changed or created, on a large catalog, we might prefer only
using the cron to have less frequent updates and less jobs.

A second fix is that when a stock_location_id is set on the backend, it
should be used, it was only used for the import of stock.

### DIFF
--- a/connector_prestashop/models/product_template/common.py
+++ b/connector_prestashop/models/product_template/common.py
@@ -3,7 +3,7 @@
 
 from collections import defaultdict
 
-from openerp import _, exceptions, api, fields, models
+from openerp import api, fields, models
 from openerp.addons.decimal_precision import decimal_precision as dp
 
 from ...unit.backend_adapter import GenericAdapter
@@ -111,8 +111,6 @@ class PrestashopProductTemplate(models.Model):
         digits_compute=dp.get_precision('Product Price'),
     )
 
-    RECOMPUTE_QTY_STEP = 1000
-
     @api.multi
     def recompute_prestashop_qty(self):
         # group products by backend
@@ -126,28 +124,7 @@ class PrestashopProductTemplate(models.Model):
 
     @api.multi
     def _recompute_prestashop_qty_backend(self, backend):
-        root_location = (backend.stock_location_id or
-                         backend.warehouse_id.lot_stock_id)
-        locations = self.env['stock.location'].search([
-            ('id', 'child_of', root_location.id),
-            ('prestashop_synchronized', '=', True),
-            ('usage', '=', 'internal'),
-        ])
-        # if we choosed a location but none where flagged
-        # 'prestashop_synchronized', consider we want all of them in the tree
-        if not locations:
-            locations = self.env['stock.location'].search([
-                ('id', 'child_of', root_location.id),
-                ('usage', '=', 'internal'),
-            ])
-        if not locations:
-            # we must not pass an empty location or we would have the
-            # stock for every warehouse, which is the last thing we
-            # expect
-            raise exceptions.UserError(
-                _('No internal location found to compute the product '
-                  'quantity.')
-            )
+        locations = backend._get_locations_for_stock_quantities()
         self_loc = self.with_context(location=locations.ids,
                                      compute_child=False)
         for product in self_loc:
--- a/connector_prestashop/models/product_template/common.py
+++ b/connector_prestashop/models/product_template/common.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
-from openerp import api, fields, models
+from openerp import _, exceptions, api, fields, models
 from openerp.addons.decimal_precision import decimal_precision as dp
 
 from ...unit.backend_adapter import GenericAdapter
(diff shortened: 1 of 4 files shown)


simahawk (approved):

connector_prestashop/models/product_template/common.py
```diff
@@ -118,11 +118,28 @@ def recompute_prestashop_qty(self):
...
+        if not locations:
+            locations = self.env['stock.location'].search([
+                ('id', 'child_of', root_location.id),
+                ('usage', '=', 'internal'),
+            ])
+        if not locations:
+            # we must not pass an empty location or we would have the
+            # stock for every warehouse, which is the last thing we
+            # expect
+            raise exceptions.UserError(
```
pedrobaeza: But is this going to be silently thrown to the log only or is this reflected on a connector job?

guewen: It is called either by a click on the button to update stocks, either in a job. In both cases we'll have an error poping to the UI or in the job.

Also note that it should really be a rare error since it means that you selected a location which is not internal and has no no internal children. But that's really better than the previous behavior which was to export the added stock of all warehouses.

pedrobaeza: OK

pedrobaeza:
Travis error is again the intermittent problem?

simahawk:
@pedrobaeza yes, passed on OCB failed on ODOO this time

simahawk:
@pedrobaeza @sergio-teruel could we merge this?

simahawk:
@pedrobaeza same here :stuck_out_tongue:

pedrobaeza:
Please change version number also.

simahawk:
@pedrobaeza version should be ok now.

pedrobaeza:
Can I squash commits?

guewen:
@pedrobaeza the documentation of the individual changes is in the commits, if you squash, it's harder to know the reasons why I did a change later

pedrobaeza:
I don't see special importance to keep middle algorithms that are not the final ones and documentation should be better as comments before the code lines. Also a commit that switches the module number is not relevant to be kept IMO.

guewen:
Example: in commit 22926c7 there is this implementation:

```
backends = defaultdict(self.browse)
for product in self:
    backends[product.backend_id] |= product
```

In 5d617fd I change it to 

```
backends = defaultdict(set)
for product in self:
    backends[product.backend_id].add(product.id)
```

If in some days, someone has the bad idea to say, hey it's more cool to use `backends = defaultdict(self.browse)` instead of a set, then there is a documentation of why this is not something to do and that it has been done in the past.

A commit must explain why you change something. This is not a comment in the code that will explain why you make a change.

Regarding the version number, this would not be cleaner to choose one of the commit to melt it inside, but I would argue that: normally a version number should not be changed in the PR, but by a maintainer directly on the main branch when needed. But well, nobody does that here.

pedrobaeza:
@guewen thanks for your explanations, but I don't agree: if I change the implementation, I prefer to have in the code the explanation:
```python
# Using sets instead of recordsets because they are slow
# see https://github.com/guewen/connector-magento/pull/9
backends = defaultdict(set)
for product in self:
    backends[product.backend_id].add(product.id)
```
For sure most of the developers won't look for previous implementations of this (and more when there's a lot of commits and Git is not smart enough to show you with "selection history" the previous lines). Having this comment, you won't doubt.

About this particular optimization (I haven't made a proper review, just helping you in merging this), I'm curious as I'm not really sure about the performance improvement. Recordsets use set operations with ids in the background, so there shouldn't be too much performance penalty. Is this something you have checked (specially on v9/v10)?

guewen:
> About this particular optimization (I haven't made a proper review, just helping you in merging this), I'm curious as I'm not really sure about the performance improvement. Recordsets use set operations with ids in the background, so there shouldn't be too much performance penalty. Is this something you have checked (specially on v9/v10)?

That was my thought too before comparing the 2. @Cedric-Pigeon actually found that. Can take dozen of minutes for thousands of records in with browse vs seconds with sets.

pedrobaeza:
Thanks for the explanation, I merge although the disagreement on the commit policy, but please take into account my comments (and way of thinking) for future developments.

guewen:
> but please take into account my comments (and way of thinking) for future developments.

I respect your way of thinking which does not mean I agree. **Commits tell a story** and I'm far to be the only one into this line of thinking. When debugging and searching the reason for anything we see in the code, atomic and explained commit messages are invaluable. Which doesn't mean comments in the code are not, they are complementary. I won't start to throw all my code in big fat commits just because you prefer to. 

Thanks for the merge.

pedrobaeza:
I'm not saying that. I'm just asking to be balanced between both. You tend to go only for the commit side.

PR #63, opened 2017-06-26
8.0 - Fix import product translations

hugosantosred:
Call super in product_template _after_import to import product translations.


liebana:
:+1:

cc @jcarlosmontoya @JGarcia-Panach @sergio-teruel

PR #71, opened 2017-08-17
[fix] add missing `base_vat` dependency

simahawk:
Partner address importer relies on `simple_vat_check`
to validate VAT and drop a checkpoint if needed.
The feature comes from `base_vat` which could be
not installed (most localizations requires it, not all).

Prevents:
```
  File "[...]/connector-prestashop/connector_prestashop/models/res_partner/importer.py", line 205, in _check_vat
    return partner_model.simple_vat_check(vat_country, vat_number)
AttributeError: 'res.partner' object has no attribute 'simple_vat_check'
```

### DIFF
--- a/connector_prestashop/__openerp__.py
+++ b/connector_prestashop/__openerp__.py
@@ -11,6 +11,7 @@
     "license": "AGPL-3",
     "depends": [
         "account",
+        "base_vat",  # for vat validation on partner address
         "product",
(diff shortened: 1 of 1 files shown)


simahawk:
travis failing as usual for the export qty flappy job.... :cry:

guewen (approved):

pedrobaeza (approved):

PR #72, opened 2017-08-18
Fix path to id

guewen:
'value' is a product.attribute.value, which is the type of record
expected in 'odoo_id', so there is no such thing as 'val_id'


simahawk (approved):

simahawk:
one of the test fails as usual for export qty flappy test

simahawk (changes requested):
@guewen can you increase version number?

simahawk (approved):

simahawk:
this one has been fixed into #82

pedrobaeza:
Superseeded by #82

PR #80, opened 2017-09-01
[add] default lang to backend and fix lang for write

simahawk:


connector_prestashop/models/prestashop_backend/common.py
```diff
@@ -95,6 +95,11 @@ def _select_versions(self):
         inverse_name='backend_id',
         string='Languages',
     )
+    default_language_id = fields.Many2one(
+        comodel_name='prestashop.res.lang',
+        domain=[('active', '=', True)],
```
guewen: Isn't this domain implicit?

flotho:
Hi,
I've initiated a proposal for this question : https://github.com/OCA/connector/issues/282*
Maybe you could react and make some additionnal proposal?

Regards

leemannd (approved):

Issue #77, opened 2017-08-24
Export quantity job test fails randomly


simahawk:
Just tracking down an error that we are experiencing since a long time now...
Sometimes it passes, sometimes not.
All the PRs pending now are experiencing this.

![image](https://user-images.githubusercontent.com/347149/29660446-a81e7b44-88c1-11e7-9521-754be91122dc.png)

PR #81, opened 2017-09-08
Fix atomicity design

nbessi:
That may pass the prestashop id instead of the binding id to the exporter
That can cause the following problems:

* A job Error: Key (id_attribute_group)=(4) is not present in table "prestashop_product_combination_option"

* Wrong combination or combination values export or updated
* Combination or combination value never exported
* Faulty value in fields 'id_attribute_group' leading to data
  incoherence

### DIFF
--- a/connector_prestashop/__openerp__.py
+++ b/connector_prestashop/__openerp__.py
@@ -7,7 +7,7 @@
 
 {
     "name": "PrestaShop-Odoo connector",
-    "version": "9.0.1.0.5",
+    "version": "9.0.1.0.6",
     "license": "AGPL-3",
     "depends": [
         "account",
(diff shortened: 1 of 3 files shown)


nbessi:
Still WIP. I must test with a real setup an ensure test are green.
This should probably be ported to version 10

nbessi:
fixes #77 and probably other pr

simahawk:
@sergio-teruel @pedrobaeza can you have a look at this? 
It fixes a big issue w/ combination sync and - very likely - the flapping test.

simahawk:
I triggered a couple of times the rebuild of runbot job... not sure way it's always marked as "skipped"

pedrobaeza:
Please change module version number

nbessi:
@pedrobaeza done

PR #82, opened 2017-09-13
Improve and fix qty handling

simahawk:
* make qty field configurable via backend
* unify qty computation within template and variant as the code was really the same
* fix to avoid negative qty pushed to PS

### DIFF
--- a/connector_prestashop/__openerp__.py
+++ b/connector_prestashop/__openerp__.py
@@ -20,7 +20,7 @@
         "product_variant_supplierinfo",  # oca/product-variant
         # TODO: perhaps not needed:
         # "product_variant_cost_price",  # oca/product-variant
-        'stock_available_immediately', # oca/stock-logistics-warehouse
+        'stock_available_immediately',  # oca/stock-logistics-warehouse
     ],
     "external_dependencies": {
         'python': [
--- a/connector_prestashop/__openerp__.py
+++ b/connector_prestashop/__openerp__.py
@@ -20,6 +20,7 @@
         "product_variant_supplierinfo",  # oca/product-variant
         # TODO: perhaps not needed:
         # "product_variant_cost_price",  # oca/product-variant
+        'stock_available_immediately', # oca/stock-logistics-warehouse
     ],
     "external_dependencies": {
         'python': [
--- a/connector_prestashop/models/prestashop_backend/common.py
+++ b/connector_prestashop/models/prestashop_backend/common.py
@@ -122,6 +122,16 @@ class PrestashopBackend(models.Model):
     )
     verbose = fields.Boolean(help="Output requests details in the logs")
     debug = fields.Boolean(help="Activate PrestaShop's webservice debug mode")
+    product_qty_field = fields.Selection(
+        selection=[
+            ('immediately_usable_qty', 'Immediately usable qty'),
+            ('qty_available', 'Qty available'),
+        ],
+        string='Product qty',
+        help='Select how you want to calculate the qty to push to PS. ',
+        default='qty_available',
+        required=True,
+    )
 
     @api.model
     def _default_pricelist_id(self):
--- a/connector_prestashop/models/product_product/common.py
+++ b/connector_prestashop/models/product_product/common.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from collections import defaultdict
 from openerp import api, fields, models
 from openerp.addons.decimal_precision import decimal_precision as dp
 
@@ -139,7 +138,10 @@ class ProductProduct(models.Model):
 
 class PrestashopProductCombination(models.Model):
     _name = 'prestashop.product.combination'
-    _inherit = 'prestashop.binding.odoo'
+    _inherit = [
+        'prestashop.binding.odoo',
+        'prestashop.product.qty.mixin',
+    ]
     _inherits = {'product.product': 'odoo_id'}
 
     odoo_id = fields.Many2one(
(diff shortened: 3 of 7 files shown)


connector_prestashop/models/prestashop_backend/common.py
```diff
@@ -120,6 +120,16 @@ def _select_versions(self):
         string='Importable sale order states',
         help="If valued only orders matching these states will be imported.",
     )
+    product_qty_field = fields.Selection(
+        selection=[
+            ('immediately_usable_qty', 'Immediately usable qty'),
+            ('qty_available', 'Qty available'),
```
guewen: `virtual_available` is a good choice too (IMO way better choice than `qty_available`, I fail to see the use case where you want to sell the `qty_available` without consideration for the outgoing quantities). `immediately_usable_qty` is only possible if the module `stock_available_immediately` is installed, so should we add this option conditionally or be less restrictive and allow user the choose any field (including obsviously wrong fields) like in the magento connector: https://github.com/OCA/connector-magento/blob/66db2f4bcc9bff4abf75ce684a31d695bf295bc4/magentoerpconnect/magento_model.py#L153-L162

sergio-teruel: Hi.. We could do PS dependent of this module, I think that incoming qty should not be exported to PS.. 
That you seem to add this dependency??

guewen: I'd agree with adding the dependency as indeed it seems that immediately_usable_qty is the most sensible choice.

simahawk: `qty_available` is the current one and I kept it as I didn't want to screw existing installations that might depends on it. Not sure if we should drop it now or if we should just log its deprecation.

+1 for dependency.

guewen (approved):

guewen:
Only lint is failing with:

connector_prestashop/__openerp__.py:23:39: E261 at least two spaces before inline comment

simahawk:
rebased and fixed PEP8 errors (spotted some on catalog manager too)

connector_prestashop/__openerp__.py
```diff
@@ -20,6 +20,7 @@
         "product_variant_supplierinfo",  # oca/product-variant
         # TODO: perhaps not needed:
         # "product_variant_cost_price",  # oca/product-variant
+        'stock_available_immediately',  # oca/stock-logistics-warehouse
```
pedrobaeza, line 23: Why this new dependency? We shouldn't add dependencies on a stable branch

florian-dacosta:
@simahawk 
I also think these changes are very important and even essential, as it does not make really sense to send qty_available to prestashop anyway.
But ,I don't think it is reliable to continue computing the prestashop stock at quant level (create/write/unlink).
It should be done at move level, like it was in version 7. Indeed, when you confirm a stock move, it can change the outgoing qty, even without any modification on the quant.
So I think we should change the `update_prestashop_qty` part from the quant to the stock moves.
We should recompute on stock move : 
- action_confirm
- action_cancel
- action_done
If any of the location or location_dest is synchronized with presta.

It should work for any type of stock at the contrary of the stock quant. 
For example, if we take virtual qty, il should also recompute after an action on stock move...

About the dependance, I guess we could maybe remove, as the field is just a selection field. And then raise if the user choose this field and the module is not installed? with a python constraint on the field or something?
It would avoid to create a micro submodule with just an extension of the qty field and a dependance...

What do you think?

simahawk:
@florian-dacosta seems legit.
Would you mind to take over this PR to move it fwd? 
I'm not working w/ v9 nor PS since a lot and I won't get back to it in the next future.
You can simply get rid of the commit w/ the dependency addition. The others shall be good.

BTW we are using this in prod since more than one year and seems to work smoothly, at least of our customer's inventory.

pedrobaeza:
I don't think it's a good move, as stock.move are not an stock levels handler. What is the problem with quants? I don't see that problem...

florian-dacosta:
@pedrobaeza 
The problem is, we want to be able to choose which stock field to export to prestashop between qty_available, immediatly_usable_qty and virtual_qty (and maybe others, why not...)

And, while the qty_available is based only on stock_quant, the other 2 (immediatly_usable_qty and virtual_qty) are based on stock_quant and stock move.
Indead, if I create an outgoing picking, the immediatly_usable_qty changes, even if there was no change on any quant.
AFAIK, stock quant only change after an action on stock move, it will also work with this.
To see the problem, you can make a simple example : 
- import a new product from prestashop (or create one and export it, it does not matter)
- update the qty to 5. The virtual_qty should be 5 and the prestashop qty should be 5.
- Create a PO with 2 product and confirm it : the virtual_qty should be 7 and the prestashop qty is still 5.
Here is the problem.
It is the same with immediatly_usable_qty. It is less worse, because usually, the outgoing_qty is reserved, which trigger a write on stock quant, which should update the stock, but it is absolutly not 100% reliable, as you also may have case where you can't or  do not want to reserve immediatly...

@simahawk I am not working with prestashop in version 9 either actually, but I am going to migrate it to version 12, so want to make this changes in this version, at least...

pedrobaeza:
immediatly_usable_qty can be also deducted from quants - reservations, and the other one should be computed the same way as Odoo does (available qty +/- expected from moves), so again I insist in not reinventing ways of computing things, and more updating values on stock.move state changes. For me the last criteria (virtual_qty) is not usable at all in an e-commerce, as it makes to the website some quantity that are promised and you don't know when it will arrive.

And if you are going to do this in v12, things are totally different there.

florian-dacosta:
Well, I strongly disagree.
immediatly_usable_qty in the module `stock_available_immediately` is not deducted from quant reservations. Like you can see in the example I made in my last comment, if we based ourselves on the  quant, sometimes we are wrong. 
For the virtual qty, it is even worse, the stock quant are really far from enought to trigger the recompute at the good moment.

I am not trying to reinvent anything, I let Odoo compute the stock, I just want to change the trigger that send the qty to prestashop, because the actual one is not 100 reliable.

You may think the virtual qty is not usable in e-commerce, other may think differently, the solution I propose work for all kind of qty...

Finally, I don't see how it is totally different in v12, the qty_available, virtual_qty or immediatly_usable_qty are still computed the same way and depend of stock quant and stock move. The problem is the same in version 12... Unless I missed something important, in that case, I would appreciate an explanation.

And finally, I think if we trigger this on stock move confirmation/transfer/cancel,it won't be recomputed more than it does already, on every stock quant write/create/unlink...
So performance should not be worse.
The  code would not be more complicated, it would not add nor remove code, basically just move it. It  will just be more generic for all kind of stock qty we may want to use.

It would be nice to have other opinions about this. And it does not require to know the  connector prestashop.

PR #83, opened 2017-09-22
Raise when PS did not create a stock_availables

guewen:
We observed a bug in PrestaShop where when we create
a product, sometimes PrestaShop does not a record in the
'stock_availables' table, then we get an empty list of ids.
Fail hard so we are noticed about this issue. Otherwise,
we would just not be aware that no quantity has been sent.

### DIFF
--- a/connector_prestashop/models/product_template/common.py
+++ b/connector_prestashop/models/product_template/common.py
@@ -3,8 +3,10 @@
 
 from collections import defaultdict
 
-from openerp import api, fields, models
+from openerp import _, api, fields, models
 from openerp.addons.decimal_precision import decimal_precision as dp
+
+from openerp.addons.connector.exception import FailedJobError
 
 from ...unit.backend_adapter import GenericAdapter
 from ...backend import prestashop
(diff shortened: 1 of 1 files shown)


connector_prestashop/models/product_template/common.py
```diff
@@ -166,7 +168,20 @@ def export_quantity(self, filters, quantity):
     def export_quantity_url(self, filters, quantity, client=None):
         if client is None:
             client = self.client
+        # Get the ids of the 'stock_availables' records for this product.
+        # Thereafter, we will update those records with the new quantities.
         response = client.search(self._prestashop_model, filters)
+        if not response:
```
pedrobaeza: Can it be that there's response, but with empty list as stated in the main comment?

pedrobaeza:
Please change module version number

connector_prestashop/models/product_template/common.py
guewen: Here `responses` is the list of ids, `search` returning a list of ids, the variable name is a bit confusing in that regard (should I rename it?)

pedrobaeza: Well, for sure we will ask the same in the future, so if you don't mind, it would be best, but I won't block it for that reason. A version number changed would also be desirable, but as there are other PRs from @simahawk, I don't want to be again a conflictive change, so I will merge all together so that an effective version change happens.

guewen: Agreed, it's renamed. As for the version number, yeah... the conflict would be an issue :-(
Thanks!

leemannd:
@guewen Increase version number added in https://github.com/guewen/connector-prestashop/pull/16

PR #93, opened 2018-05-16
Prestashop 1.7 compatibility

metaminux:
Hello,

I'm trying to use this connector with Prestashop 1.7 (1.7.1.1 actually)

The first thing I had to change is setting the field "state" which is new in 1.7 on model Products in order to get exported products from Odoo visible in Prestashop backoffice.

There should be other changes needed...


connector_prestashop_catalog_manager/__openerp__.py
```diff
@@ -3,17 +3,19 @@
...
     "name": "Prestashop-Odoo Catalog Manager",
-    "version": "8.0.1.0.2",
+    "version": "8.0.1.0.3",
     "license": "AGPL-3",
     "depends": [
         "connector_prestashop"
     ],
     "author": "Akretion,"
               "AvanzOSC,"
+              "Meta IT,"
```
pedrobaeza: For this little contribution you shouldn't be in author list, but as contributor in the README

pedrobaeza:
You should add also the field to the view

metaminux:
@sergio-teruel An update on this PR ?

Thanks

FFernandez-PlanetaTIC (approved):

FFernandez-PlanetaTIC (commented):
Does PS 1.6 work with new mapping `state`?

pedrobaeza:
Why does this provide compatibility with 1.7? I only see a new fields that is mapped, but it's not an compatibility layer. I also see translations that must be done on Weblate for avoiding conflicts on an already merged module: https://odoo-community.org/page/translate

FFernandez-PlanetaTIC:
PS v1.7 adds `state` field for logical deletion. Products created from Odoo are not seen in PS if this field is not set. It fixes v1.7 but I think it can break v1.6 because this fields does not exist.

pedrobaeza:
Yes, I think so. A different mapper must be done.

pedrobaeza:
And translations must go out.

FFernandez-PlanetaTIC (changes requested):
Do new mapper for v1.7 state field and take out translations

FFernandez-PlanetaTIC:
I just tested the mapping `state` with version 1.6.1.6 and it has not been broken.

FFernandez-PlanetaTIC (changes requested):
Take out translations.

metaminux:
@FFernandez-PlanetaTIC Like suggested, I took the translations out...

FFernandez-PlanetaTIC (approved):

FFernandez-PlanetaTIC:
@metaminux Thanks

pedrobaeza (approved):

PR #95, opened 2018-09-21
[9.0] [FIX] catalog_manager: Fix images export

FFernandez-PlanetaTIC:
Hi,

It seems that the export of images does not work. This fix that.

Regards.

cc @PlanetaTIC

### DIFF
--- a/connector_prestashop_catalog_manager/models/product_product/common.py
+++ b/connector_prestashop_catalog_manager/models/product_product/common.py
@@ -5,6 +5,14 @@
 from openerp import models, fields
 
 
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    image_ids = fields.One2many(
+        inverse=False,
+    )
+
+
 class PrestashopProductCombination(models.Model):
     _inherit = 'prestashop.product.combination'
     minimal_quantity = fields.Integer(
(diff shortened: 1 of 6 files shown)


pedrobaeza:
@sergio-teruel is that true?

connector_prestashop_catalog_manager/models/product_template/common.py
```diff
@@ -2,10 +2,24 @@
...
-from openerp import models, fields
+from openerp import api, models, fields
 import openerp.addons.decimal_precision as dp
 
 
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    @api.multi
+    def _set_multi_image_main_medium(self):
```
FFernandez-PlanetaTIC, line 13: The main image of a product is exported three times when it's modified:

jobs

This is due to the design of module `product_multi_image`.

Each export involves the removal and creation again of the image in Prestashop and concurrent exports can fail and break the binding.

The export of medium and small images can be avoided.

connector_prestashop_catalog_manager/models/product_product/common.py
```diff
@@ -5,6 +5,14 @@
 from openerp import models, fields
 
 
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    image_ids = fields.One2many(
+        inverse=False,
```
FFernandez-PlanetaTIC, line 12: A combination image is exported several times when it's modified:

jobs

This is a weird behavior of One2Many inverse and write listener, which is fired once per image field.

The modification of images has been disabled from the combinations to avoid it.

FFernandez-PlanetaTIC:
I've also added a couple of commits here to avoid unnecessary export of images and save concurrent failures and binding breaks.

PR #100, opened 2018-10-02
10 imp flotho

bguillot:
This is the last commits that haven't been merge in Technativa branch.

PR #101, opened 2018-10-08
[10.0][MIG]Add rest of Akretion modifications

flotho:
Some part of PR haven't been pushed to Akretion before OCA merge @bguillot  PR.
This PR propose to re-add all missing parts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,5 @@ coverage.xml
 # Sphinx documentation
 docs/_build/
 
+#OS directory manager
+.directory
--- a/connector_prestashop/README.rst
+++ b/connector_prestashop/README.rst
@@ -8,6 +8,9 @@ Odoo PrestaShop Connector
 
 This module connects Odoo and PrestaShop.
 
+Caution : it's not compatible with product_variant_configurator
+(see models/produc_product/common.py)
+
 PrestaShop (http://www.prestashop.com/) is a popular e-commerce platform
 written in PHP/MySQL and published under the Open Software licence v3.0.
 
--- a/connector_prestashop/backend.py
+++ b/connector_prestashop/backend.py
@@ -13,3 +13,5 @@ prestashop_1_6_0_9 = backend.Backend(parent=prestashop, version='1.6.0.9')
 prestashop_1_6_0_11 = backend.Backend(parent=prestashop, version='1.6.0.11')
 # version >= 1.6.1.2
 prestashop_1_6_1_2 = backend.Backend(parent=prestashop, version='1.6.1.2')
+# version = 1.7.5.0
+prestashop_1_7_5_0 = backend.Backend(parent=prestashop, version='1.7.5.0')
--- a/connector_prestashop/components/auto_matching_importer.py
+++ b/connector_prestashop/components/auto_matching_importer.py
@@ -74,7 +74,7 @@ class AutoMatchingImporter(Component):
                             'backend_id': self.backend_record.id,
                         }
                         for oe_field, ps_field in self._copy_fields:
-                            data[oe_field] = ps_dict[ps_field]
+                            data[oe_field] = erp_dict[ps_field]
                         record = self.model.create(data)
                         binder.bind(ps_id, record)
                         _logger.debug(
--- a/connector_prestashop/components/importer.py
+++ b/connector_prestashop/components/importer.py
@@ -507,10 +507,13 @@ class TranslatableRecordImporter(AbstractComponent):
         super(TranslatableRecordImporter, self)._import(binding)
 
     def _after_import(self, binding):
+        # TODO: Improve the way translations are called
+        # https://github.com/OCA/connector/issues/282
         """ Hook called at the end of the import """
         for lang_code, lang_record in self.other_langs_data.iteritems():
             map_record = self.mapper.map_record(lang_record)
             binding.with_context(
                 lang=lang_code,
                 connector_no_export=True,
+                force_translate=True,
             ).write(map_record.values())
--- a/connector_prestashop/components/importer.py
+++ b/connector_prestashop/components/importer.py
@@ -379,7 +379,7 @@ class DelayedBatchImporter(AbstractComponent):
             description=description,
             channel=channel,
             identity_key=identity_key
-            ).import_record(
+        ).import_record(
             backend=self.backend_record,
             prestashop_id=external_id,
             **kwargs)
--- a/connector_prestashop/components/version_key.py
+++ b/connector_prestashop/components/version_key.py
@@ -8,6 +8,7 @@ import logging
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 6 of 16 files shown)


flotho:
Replace :
* https://github.com/OCA/connector-prestashop/pull/100
* https://github.com/akretion/connector-prestashop/pull/7

pedrobaeza:
Can you check Travis?

gaelTorrecillas:
hello @oca-clabot, 
I upgraded my account.

Regards,

pedrobaeza:
Travis is still red. Can you please check?

flotho:
Hi @pedrobaeza , are we almost there?
Are the tests failure normal regarding to you ?

regards

FFernandez-PlanetaTIC:
Hi @flotho ,

We have the same test failure in #103 and we are reviewing it.

Regards!

FFernandez-PlanetaTIC:
Hello again,

Our #104 is also affected by the test failure.

We don't get this error in our environments and I don't understand why... :confused: 

I think it's related to #77 , which should have been fixed with #81 .

Regards

flotho:
Hi @FFernandez-PlanetaTIC , it's also failing randomly in our tests.
@pedrobaeza any chance to have a positive review for a merge or do we have to also deal with this issue?
regards

pedrobaeza:
The idea of tests is to have predictable results, so it's weird that this happens randomly. There should be a cause. Isn't possible to debug what is the underlying problem and fix it?

flotho:
Hi @pedrobaeza , of course it should be possible.
Yet either @FFernandez-PlanetaTIC and our team have some difficulties to reproduce.
Moreover, many proposal for v10 are in the pipe.
I'm not sure we'll be able to find what's happening soon. @FFernandez-PlanetaTIC , what's your point of view?

FFernandez-PlanetaTIC:
I have debugged the failing test on friday and today but it doesn't fail in our environments and I can't find the cause :confused:

Maybe @simahawk and @nbessi can give us a clue as it seems that they had the same problem in #77 and #81 .

FFernandez-PlanetaTIC:
Finally we reproduced and fixed it in #104 

Damn cache! :rage:

pedrobaeza:
@FFernandez-PlanetaTIC 2 things about this:

* Please do a separate PR for fixing Travis (and possibly fast-tracking).
* About the solution, this is revealing a bad `depends` declaration on any of the fields that you are handling. This fixes tests, but probably in real world you would face the same problem. Can you dig a bit more which field contains the incorrect value and then deduce the dependent fields that are not declared?

FFernandez-PlanetaTIC:
Hi @pedrobaeza ,

I think it's a problem in _stock_ module, a quick test:

```
paco@coloma$ bin/python_odoo 
To start the Odoo working session, just do:
    session.open(db=DATABASE_NAME)
or, to use the database from the buildout part config:
    session.open()
All other options from buildout part config do apply.
Then you can issue commands such as:
    session.registry('res.users').browse(session.cr, 1, 1)
Or using new api:
    session.env['res.users'].browse(1)

>>> session.open('test_10.0_catalog_manager')
2018-12-18 10:08:59,791 20350 INFO test_10.0_catalog_manager odoo.modules.loading: loading 1 modules...
2018-12-18 10:08:59,862 20350 INFO test_10.0_catalog_manager odoo.modules.loading: 1 modules loaded in 0.07s, 0 queries
2018-12-18 10:08:59,898 20350 INFO test_10.0_catalog_manager odoo.modules.loading: loading 64 modules...
2018-12-18 10:09:00,003 20350 INFO test_10.0_catalog_manager odoo.addons.report.models.report: Will use the Wkhtmltopdf binary at /usr/local/bin/wkhtmltopdf
2018-12-18 10:09:00,295 20350 INFO test_10.0_catalog_manager odoo.modules.loading: 64 modules loaded in 0.40s, 0 queries
2018-12-18 10:09:00,772 20350 INFO test_10.0_catalog_manager odoo.modules.loading: Modules loaded.
>>> 
>>> product = session.env['product.product'].create({'name': 'Test'})
>>> product.qty_available
0.0
>>> product.product_tmpl_id.qty_available
0.0
>>> 
>>> warehouse = session.env['stock.warehouse'].search([], limit=1)
>>> 
>>> qty_change = session.env['stock.change.product.qty'].create({'location_id': warehouse.lot_stock_id.id, 'product_id': product.id, 'new_quantity': 10})
>>> qty_change.change_product_qty()
{'type': 'ir.actions.act_window_close'}
>>> 
>>> product.qty_available
10.0
>>> product.product_tmpl_id.qty_available
0.0
>>> 
>>> session.env.invalidate_all()
>>> product.product_tmpl_id.qty_available
10.0
```

What do you think? Should we create an issue in odoo/odoo or OCA/OCB?

Thanks!

pedrobaeza:
Thanks for the insights, @FFernandez-PlanetaTIC. That has been very helpful!

Indeed, there's a problem in Odoo code. Look the compute method of `product.product`.`qty_available` field:

https://github.com/odoo/odoo/blob/10.0/addons/stock/models/product.py#L100-L101

It has their corresponding depending fields for properly recomputing things when one of the values change.

But for `product.template`, what we have is:

https://github.com/odoo/odoo/blob/10.0/addons/stock/models/product.py#L434-L435

without `@api.depends`, and this is what is causing the glitch. If the value hasn't been accessed previously, then it computes the right value when needed, but on contrary if it has been computed previously, then accessing again when conditions have changed will show the incorrect value.

This should be fixed at Odoo/OCB level. The solution should be - without testing further - to add these depends on the `product.template` field:

```python
@api.depends(
    'product_variant_ids',
    'product_variant_ids.stock_quant_ids',
    'product_variant_ids.stock_move_ids',
)
def _compute_quantities(self):
    ...
```

It would be great if you try to put this in Odoo (directly a PR), with the same shell console commands for illustrating the problem, and with this solution. I will ping proper people for reviewing it and hopefully to get merged. If not, we can go to OCB, but we have to check if this is also happening on v11/v12 for giving this more importance to be pushed in core.

Finally, you don't have to wait all of this for solving the problem without invalidating the cache in this module, as this only hides the problem in the tests, but in the real world can happen. This is the great thing about modularity! Just add an inherit to `product.template` with the following:
```python
class ProductTemplate(models.Model):
    _inherit = 'product.template'

    @api.depends(
        'product_variant_ids',
        'product_variant_ids.stock_quant_ids',
        'product_variant_ids.stock_move_ids',
    )
    def _compute_quantities(self):
        return super(ProductTemplate, self)._compute_quantities()
```
with this, you are adding the dependencies and achieving the same result. However, put a comment on this code for noting that this is patch until PR X gets merged in Odoo.

FFernandez-PlanetaTIC:
We have already created/modified the PRs as suggested:

- odoo/odoo#30024
- odoo/odoo#30026
- #106

PR #111, opened 2019-03-20
[WIP] Connector prestashop to version 12

ljimenezsidoo:
We can not able to do format-path because a lot of errors.

@florian-dacosta If you want to check where we are right now.


pedrobaeza:
Not valid without commit history, as you know.

florian-dacosta:
@acm1pt-colorado 
With only one commit, it is impossible to review.
Also, I have a complete working branch. We should take the one more advanced and try to complete it with the eventual missing features from the less advanced branch.

pedrobaeza:
I will do the commit rescue today. Closing this for now and waiting the PR based on my branch with the commit history.

rvalyi:
@pedrobaeza I will be interrested in the step you took roughly to re-do the branch as I think we had such an issue when trying to integrate a third party module in OCA/l10n-brazil and I believe git filter-branch was failing because they made such wild merges affecting several modules.

pedrobaeza:
The hard way: I'm already 1 hour resolving conflicts one by one and doing `git am --continue`. There's no other way, just cut the hand of whom have mangled commit history doing a `git pull`/`git merge` operation ruining the sequentially order of the commits :stuck_out_tongue_closed_eyes:

PR #112, opened 2019-03-22
[12.0][MIG] connector_prestashop: Migration to 12.0

florian-dacosta:
The first commit correspond to standard migration.
Then I made some improvements that seem necessary and generic and did one separated commit for each improvement, so it can be reviewed easier.

I believe it is ready for review and not work in progress, although, some dependencies are still missing...
Indeed, I did not have the time to migrate all dependencies and used some module in version 11 in order to have the connector running in version 12...
So, even if it is not WIP, it should not be merged before other dependencies are also migrated...

Depends on : 
https://github.com/OCA/server-tools/pull/1528
https://github.com/OCA/connector-ecommerce/pull/49


angelmoya:
Hi @florian-dacosta , what about prestapyt version? did you check on python3?

We did also migration to v12, with python3, and we are using prestatpyt3 instead of prestapyt.

We are in progress because we are working with prestashop version 1.7.4.4.

We plan to go next month to OCA Spanish Sprint Code, maybe we can improve this together.

pedrobaeza:
prestatpyt3? https://github.com/prestapyt/prestapyt/pull/48 already gave Python 3 compatibility to prestapyt

angelmoya:
Thanks @pedrobaeza  I will check

angelmoya:
OK, pip install version 0.8... install from git repository works right.

florian-dacosta:
@angelmoya 
Indeed, I use the prestapyt version indicated by Pedro.
Sadly, I won't be able to go to the code sprint in Spain at the end of May...
I am going to one organised in Berlin in 2 weeks though, but I guess you would not be there.
Anyway I would have no problem using your version instead of mine if needed, in that case, I will just need you to take into account some improvment I made. It is not so much code to review if you ignore the diff coming from the tests.
And of course, compatibility with version 1.7.4.4 will be welcome.

angelmoya:
Nice @florian-dacosta I'm checking your version, maybe I can create a PR to your branch with our improvements for this version, I will check my agenda for next week maybe I can go at lest one day.

florian-dacosta:
Yes, you can create a PR on top of mine, I'll do my best to be available and review to make things move forward.

florian-dacosta:
As the historic of the prestashop_connector has been mixed up a lot, I have no idea of who implemented the "impact_price" notion on variants. And it seems it comes from version 8, that is why I try to ping the major contributors since that time. @pedrobaeza @guewen @sergio-teruel @flotho @simahawk @angelmoya 
Any ideas about this?

We can see what I talk about here for instance : https://github.com/OCA/connector-prestashop/blob/10.0/connector_prestashop/models/product_product/common.py#L62
The impact_price notion is added and seems to replace the price_extra native odoo notion.
I wonder why? Indeed, it seems very dangerous as the price extra is used a lot in Odoo and it seems as very light implementation.
I may be wrong but I'd like the opinion on people who worked on this, in order to avoid reverting something important.

Anyway, I was doing some test on version 12 and had some big surprises about this, as it indeed is not compatible at all with some Odoo features, like pricelists for instance!
I guess it is because in prestashop the price extra notion is not really linked to the attribute, but only at the variant, unlike Odoo. I try to fix all this in my last commit, I'll do more tests, but I am not a big expert on the Odoo variant system...

pedrobaeza:
The question is that the pricing philosophy of Odoo is incompatible with PS pricing one: while Odoo gets the variant price through template base price + sum of all the price extra of each attribute value that has one, PrestaShop gets the variant price through the base price of the template + impact price for that specific variant. You can then easily deduce how incompatible gets this in the moment you introduce incoherent impact prices in PrestaShop that forbids to deduce what can be the possible extra prices for the attribute values. That's why the only solution is to put a new field totally disconnected from the pricing system of Odoo. We didn't need to make this price work on Odoo backend, as all the sales orders come from PrestaShop, and thus, we didn't connect this new field with anything. If you need it, you can overwrite at product.product level the `lst_price` compute method and replace current logic by `list_price` + `impact_price`.

florian-dacosta:
@pedrobaeza Thanks, that's what I thought.
Did you have a look to the last commit of this branch?
I think it resolve the problem a bit better than overriding the lst_price.
I guess I'll find more problem of compatibility over time but I can fix it litlle by little.

connector_prestashop/models/account_invoice/importer.py
```diff
@@ -0,0 +1,244 @@
...
+        record = self.prestashop_record
+        self._import_dependency(
+            record['id_customer'], 'prestashop.res.partner'
+        )
+        self._import_dependency(record['id_order'], 'prestashop.sale.order')
+
+    def _open_refund(self, binding):
+        invoice = binding.odoo_id
+        if invoice.amount_total == float(self.prestashop_record['amount']):
+            invoice.signal_workflow('invoice_open')
```
ivantodorovich:
```suggestion
            invoice.action_invoice_open()
```

connector_prestashop/models/account_invoice/importer.py
```diff
@@ -0,0 +1,244 @@
...
+        if price_unit in [0.0, '0.00']:
+            return None
+        product = self.env['product.product'].browse(
+            order_line['product_id'][0]
+        )
+        account_id = product.property_account_income_id.id
+        if not account_id:
+            account_id = product.categ_id.property_account_income_categ_id.id
+        if fpos:
+            fpos_obj = self.env['account.fiscal.position']
```
ivantodorovich: image

and then


image

connector_prestashop/models/account_invoice/importer.py
```diff
@@ -0,0 +1,244 @@
...
+                self.session.cr,
+                self.session.uid,
+                fpos,
+                account_id
+            )
+        return {
+            'quantity': 1,
+            'product_id': product.id,
+            'name': order_line['name'],
+            'invoice_line_tax_id': [(6, 0, order_line['tax_id'])],
```
ivantodorovich:
```suggestion
            'invoice_line_tax_ids': [(6, 0, order_line['tax_id'])],
```

https://github.com/odoo/odoo/blob/12.0/addons/account/models/account_invoice.py#L1688

I don't know when it was renamed, but it has been a while now. My guess is previous versions have this issue too

connector_prestashop/models/account_invoice/importer.py
```diff
@@ -0,0 +1,244 @@
...
+            product_id = product.id
+            name = order_line.name
+            for tax in order_line.tax_id:
+                tax_ids.append(tax.id)
+            account_id = product.property_account_income_id.id
+            if not account_id:
+                categ = product.categ_id
+                account_id = categ.property_account_income_categ_id.id
+        if fpos and account_id:
+            fpos_obj = self.session.pool['account.fiscal.position']
```
ivantodorovich: Same comment than before.

If you replace `account_id` for `account`, and then use `account.id` in the return dict
```py
        if fpos and account:
            account = fpos.map_account(account)
```

connector_prestashop/models/account_invoice/importer.py
```diff
@@ -0,0 +1,244 @@
...
+
+        discount = False
+        if price_unit in ['0.00', ''] and order_line is not None:
+            price_unit = order_line['price_unit']
+            discount = order_line['discount']
+        return {
+            'quantity': quantity,
+            'product_id': product_id,
+            'name': name,
+            'invoice_line_tax_id': [(6, 0, tax_ids)],
```
ivantodorovich: Same comment than before
```suggestion
            'invoice_line_tax_ids': [(6, 0, tax_ids)],
```

ivantodorovich (changes requested):

florian-dacosta:
@ivantodorovich Thanks for the suggestions, I added them

ivantodorovich (approved):
Great! 👍 thanks it looks very good

pedrobaeza:
Please check Travis and runbot status

florian-dacosta:
@pedrobaeza 
I made a pass on the pylint stuff, let's see if it becomes green.
But, this PR still depends on 2 others PR (I have updated the first message), so runbot won't pass for now.
This PR is not ready to merge yet.

MME83:
Hi guys, i'm newbie but i've tried this branch to install on localhost on Odoo 12v, and in dependencies we have some packages and modules from 11v like as connector-ecomerce (only 11v), sale-exeption (11v), base_multi_image (10v), sale-workflow (11v , from this package we need  sale_automatic_workflow and sale_automatic_workflow_payment_mode because in v12 they did not exist in the package sale-workflow 12), but i tried to install all depends and connected to local presta-website on my home virtualhost, the module (prestashop connector) has connected in fact, but in real no import function works properly (i think there is a problem with modules/packeges vershions or some miss tamplates)... if need i can make log files from begining of installation to working process...

ivantodorovich:
> ckages and modules from 11v like as connector-ecomerce (only 11v), sale-exeption (11v), base_multi_image (10v), sale-workflow (11v , from this package we need sale_automatic_workflow and sale_automatic_workflow_payment_mode because in v12 they did not exist in the package sale-workflow 12), but i tried to install all depends and connected to local presta-website on my home virtualhost, the module (prestashop connector) has connected in fact, but in real no import function works properly (i think there is a problem with modules/packeges vershions or some miss tamplates)... if need i can make log files from begining of installation to working process...

Hi @MME83 

This module and some of its dependencies are not yet migrated, so it's not going to work.
If you want to test it you have to use the branches of the migration PRs for every dependent module.

ie:
https://github.com/OCA/connector-ecommerce/pull/49
https://github.com/OCA/server-tools/pull/1528
and many others..

Some of these modules also depend on other modules that are still being reviewed for v12.

ivantodorovich:
Just leaving this here:
https://github.com/OCA/server-tools/pull/1528 was superseeded by https://github.com/OCA/server-tools/pull/1650

oca_dependencies.txt
```diff
@@ -0,0 +1,6 @@
+queue_job
```
ivantodorovich: repository name is `queue`

ivantodorovich (changes requested):

oca_dependencies.txt
```diff
@@ -0,0 +1,6 @@
...
+connector
+connector_ecommerce
```
ivantodorovich: connector-ecommerce

ivantodorovich (changes requested):

connector_prestashop/models/res_partner/importer.py
```diff
@@ -0,0 +1,217 @@
...
+        ('address2', 'street2'),
+        ('city', 'city'),
+        ('other', 'comment'),
+        ('phone', 'phone'),
+        ('phone_mobile', 'mobile'),
+        ('postcode', 'zip'),
+        ('date_add', 'date_add'),
+        ('date_upd', 'date_upd'),
+        ('alias', 'alias'),
+        ('company', 'company'),
```
jesusVMayor, line 138: company field don't exists in prestashop.address model

connector_prestashop/models/product_template/importer.py
```diff
@@ -0,0 +1,707 @@
...
+        if self.has_combinations(record):
+            return {}
+        code = record.get('reference')
+        if not code:
+            code = "backend_%d_product_%s" % (
+                self.backend_record.id, record['id']
+            )
+        if not self._template_code_exists(code):
+            return {'default_code': code}
+        i = 1
```
jesusVMayor, line 215: When matching_product_template is True and matching_product_ch is reference, this method changes the default_code to X_1, maybe can add:
```
if self.backend_record.matching_product_template and self.backend_record.matching_product_ch == 'reference':
            return {}
```

jesusVMayor (changes requested):
Tested on a prestashop 1.6.1, the connector works fine.

connector_prestashop/models/res_partner/importer.py
florian-dacosta: Thanks, I added it.

connector_prestashop/models/product_template/importer.py
florian-dacosta: Agree, I just made a change to fix this

connector_prestashop/models/prestashop_backend/common.py
```diff
@@ -0,0 +1,442 @@
...
+            since_date = backend_record.import_payment_mode_since
+            filters = {}
+            if since_date:
+                filters = {
+                    'date': '1', 'filter[date_upd]': '>[%s]' % (since_date)
+                }
+            with backend_record.work_on('account.payment.mode') as work:
+                importer = work.component(usage='batch.importer')
+                importer.run(filters=filters)
+            backend_record.import_payment_mode_since = now_fmt
```
florian-dacosta, line 344: I did  not get what you mean. What should also be done for other functions?
The update of the filed 'import_orders_since' and the other is already done, but in sub methods.
Did you meet an issue about this?

okuryan:
@florian-dacosta do you think this can be merged in the nearest future? What is left to be done?

florian-dacosta:
Hi @okuryan 
One dependency is not merged yet : https://github.com/OCA/server-tools/pull/1650

To move this PR forward, the module base_multi_image needs to be merged first, so reviews could be made oj base_multi_image's PR.

Then, this PR need to be reviewed also because for now, it has not been approved by anyone.

connector_prestashop/models/prestashop_backend/common.py
```diff
@@ -0,0 +1,442 @@
...
+
+    matching_customer = fields.Boolean(
+        string="Matching Customer",
+        help="The selected fields will be matched to the ref field of the "
+        "partner. Please adapt your datas consequently.")
+    # matching_customer_ch = fields.Many2one(
+    #     comodel_name='prestashop.partner.field', string="Matched field",
+    #     help="Field that will be matched.")
+
+    quantity_field = fields.Selection(
```
jesusVMayor: it seems that this field is not being used anywhere

connector_prestashop/models/prestashop_backend/common.py
```diff
@@ -0,0 +1,442 @@
...
+        "https://github.com/OCA/stock-logistics-warehouse",
+        required=True,
+        default='virtual_available')
+    tz = fields.Selection(
+        _tz_get, 'Timezone', size=64,
+        help="The timezone of the backend. Used to synchronize the sale order "
+        "date.")
+    product_qty_field = fields.Selection(
+        selection=[
+            ('immediately_usable_qty', 'Immediately usable qty'),
```
jesusVMayor: immediately_usable_qty it seems it was renamed to qty_available_not_res in stock_available_unreserved module

connector_prestashop/models/res_partner/importer.py
```diff
@@ -0,0 +1,215 @@
...
+        # the parent
+        return {'type': 'other'}
+
+
+class AddressImporter(Component):
+    _name = 'prestashop.address.importer'
+    _inherit = 'prestashop.importer'
+    _apply_on = 'prestashop.address'
+
+    def _check_vat(self, vat):
```
jesusVMayor: Maybe this function can receive the record country, and use it if the vat_number comes without the country code, like in the res.partner check_vat function.
```
def _check_vat(self, vat_number, partner_country):
        vat_country, vat_number_ = self.env["res.partner"]._split_vat(
            vat_number
        )
        if not self.env["res.partner"].simple_vat_check(vat_country, vat_number_):
            # if fails, check with country code from country
            country_code = partner_country.code
            if country_code:
                if not self.env["res.partner"].simple_vat_check(country_code.lower(), vat_number):
                    return False
        return True
```
This will reduce the number of partners who need to be checked

connector_prestashop/views/sale_view.xml
```diff
@@ -0,0 +1,155 @@
...
+            When a sale order state change, the connector will update the sale order in PrestaShop.
+          </p>
+        </field>
+    </record>
+
+    <menuitem
+        id="sale_order_state_tree_menu"
+        name="Sale Order States"
+        action="action_prestashop_sale_order_state_form"
+        parent="sales_team.menu_sale_config"
```
jesusVMayor: Maybe the parent could be sale.menu_sale_config? the sales_team menu don't have a parent, so it's render on the top, and seems ugly have this menu just for the states.

jesusVMayor (commented):
Just one problem and some improvement proposal. Otherwise, this migration seems to be wokring well in all cases that i tested. Thanks for the work @florian-dacosta

florian-dacosta:
@jesusVMayor 
Thanks for the review
Your comments seem legit to me.
I have no time to deal with it right now. If you want to, you can do a pull request on my branch and I will review and merge it fast.
If not, I'll do the necessary about your comment later (I am not sure when)

gabrielrey37:
Tested and working in odoo 12 community with prestashop 1.7.6.4
Everything works perfectly but I couldn't configure the taxes.
Somebody could help me?
for what functions in 1.7 I used this fix
https://www.prestashop.com/forums/topic/763911-ps-17-webservice-order_discounts/?tab=comments#comment-2840998

connector_prestashop/components/version_key.py
```diff
@@ -0,0 +1,71 @@
...
+        'order_row': 'order_rows',
+        'tax': 'taxes',
+        'image': 'images',
+        'combinations': 'combinations',
+        'tag': 'tags',
+        'manufacturers': 'manufacturers',
+    }
+
+
+class VersionKey1612(Component):
```
MPoch-PlanetaTIC, line 47: It would be great to add more Prestashop versions

connector_prestashop/models/product_template/importer.py
```diff
@@ -0,0 +1,708 @@
...
+            vals = {
+                'location_id': location.id,
+                'product_id': product.id,
+                'new_quantity': qty,
+            }
+            template_qty = self.env['stock.change.product.qty'].create(vals)
+            template_qty.with_context(
+                active_id=product.id,
+                connector_no_export=True,
+            ).change_product_qty()
```
MPoch-PlanetaTIC, line 491: It is not a bug, but I suggest to only call change_product when product's type is 'product'.
This way, it will not try to import stock to the products that is not required to check for the stock ('consu' and 'service' types), if not for each 'consu' and 'service' product an exception is raised.

MPoch-PlanetaTIC (commented):
I have tested this migration.
It works.

I comment two points in the code and want to remember, that to make discounts to work properly this pull request must be applied on Prestashop: https://github.com/PrestaShop/PrestaShop/pull/13393

Thank you for the migration

connector_prestashop/components/version_key.py
MPoch-PlanetaTIC: PR #104 adds Prestashop 1.6.1.12 and 1.6.1.6 versions keys

CTimmerman (approved):
Finally some actual source code.

connector_prestashop/models/account_invoice/importer.py
```diff
@@ -0,0 +1,234 @@
...
+        shipping_line = self._invoice_line_shipping(record, fpos)
+        if shipping_line:
+            lines.append((0, 0, shipping_line))
+        for slip_detail in slip_details:
+            line = self._invoice_line(slip_detail, fpos)
+            lines.append((0, 0, line))
+        return {'invoice_line_ids': lines}
+
+    def _invoice_line_shipping(self, record, fpos):
+        order_line = self._get_shipping_order_line(record)
```
MPoch-PlanetaTIC, line 109: Hello there,
A customer requested the importation of Prestashop refunds and found a little problem:
Function self._get_shipping_order_line returns a list that includes one dictionary, but here, order_line is expected to be a dictionary.
This generate an error when importing Prestashop Refunds.

I confirm that Refund Importation works correctly when solving this issue.

connector_prestashop/models/prestashop_backend/common.py
```diff
@@ -0,0 +1,431 @@
...
+        self.search(domain or []).import_products()
+
+    @api.model
+    def _scheduler_import_carriers(self, domain=None):
+        self.search(domain or []).import_carriers()
+
+    @api.model
+    def _scheduler_import_payment_methods(self, domain=None):
+        backends = self.search(domain or [])
+        backends.import_payment_methods()
```
MPoch-PlanetaTIC, line 392: Hello,
backends.import_payment_methods does not exists, this should call backends.import_payment_modes (defined above)

jesusVMayor (approved):
base_multi_image it's finally merged. We are using this code in production since a year ago. So LGTM.

PS: we don't import refunds, i don't tested the problem reported by @MPoch-PlanetaTIC

florian-dacosta:
I'll close this PR since I switch to version 14 and won't work on this anymore.
@jesusVMayor I won't delete the akretion branch now in case you or somebody else use it, but I suggest you clone it if it is not done in order to not have issue if we delete it one day.

PR #114, opened 2019-03-29
[FIX] connector_prestashop: Override external dependencies to match PyPi

### LINKED ISSUE #113: [V10] Package vcr should be vcrpy
There is no package called `vcr` on PyPI:

https://pypi.org/project/vcr/

This should be the correct one, I think:

https://pypi.org/project/vcrpy/

This issue makes the module not installable even for people who don't use tests (end users).

pedrobaeza:
Closes #113 

cc @Tecnativa

### DIFF
--- a/setup/connector_prestashop/setup.py
+++ b/setup/connector_prestashop/setup.py
@@ -2,5 +2,11 @@ import setuptools
 
 setuptools.setup(
     setup_requires=['setuptools-odoo'],
-    odoo_addon=True,
+    odoo_addon={
+        'external_dependencies_override': {
+            'python': {
+                'vcr': 'vcrpy==1.10.5',
+            },
+        },
+    },
 )


setup/connector_prestashop/setup.py
```diff
@@ -2,5 +2,11 @@
 
 setuptools.setup(
     setup_requires=['setuptools-odoo'],
-    odoo_addon=True,
+    odoo_addon={
+        'external_dependencies_override': {
+            'python': {
+                'vcr': 'vcrpy==1.10.5',
```
sbidoul, line 8: Thanks! We should avoid pinning versions here. Specifying a minimum required version is fine, while pinning must be never be done in OCA repos and done in customer projects only.

pedrobaeza: This is done on purpose as upper versions don't work.

pedrobaeza: I'm thinking that maybe we can just leave out the dependency as it's only for testing. What do you think?

sbidoul: Ah, ok, if other versions don't work then fine. It must be considered as a bug in the module though, as depending on pinned, old, versions will create problems sooner or later.

PR #123, opened 2019-07-22
[10.0] [ADD] connector_prestashop_product_dimension

FFernandez-PlanetaTIC:
Import product dimensions from PrestaShop

cc @PlanetaTIC


pedrobaeza:
As this data is integrated in product, why not do it in base module?

FFernandez-PlanetaTIC:
To avoid adding dependency with module `product_dimension`

pedrobaeza:
OK, I didn't see that... I though all the fields were in `product` module.

Issue #129, opened 2019-09-23
[10] Misleading comment for dependencies. 


adamwojt:
https://github.com/OCA/connector-prestashop/blob/af05419e40bfc9fe6a738fb6b60bde2b4cb7c1e7/connector_prestashop/__manifest__.py#L18

`base_multi_image ` comes from server-tools

pedrobaeza:
In fact I think it should be `product_multi_image`.

adamwojt:
@pedrobaeza Thank you :)

FFernandez-PlanetaTIC:
See https://github.com/akretion/connector-prestashop/pull/8#discussion_r329587052

PR #130, opened 2019-09-23
[10.0][FIX] manifest file; fixed wrong dependency

adamwojt:
As per : https://github.com/OCA/connector-prestashop/issues/129


FFernandez-PlanetaTIC:
See https://github.com/akretion/connector-prestashop/pull/8#discussion_r329587052

PR #140, opened 2020-08-12
[FIX] freeze version of freezegun

gurneyalex:
Later versions cause a crash at installation time
with obscure message:

 'module' object has no attribute 'UTC'


flotho (approved):
Hi @gurneyalex , you saved my life.
It works

flotho:
also usefull for v10

PR #152, opened 2022-05-20
[14.0]ADD keys for PS-1.7

flotho:

PR #153, opened 2022-05-21
14.0 various improvements

flotho:
Add PS 107 keys
correct bad usage of a constant
improve by adding identity key


flotho:
replace #152

flotho:
ping @saxomanu

saxomanu (approved):
LGTM

jesusVMayor (approved):

flotho:
thanks community !
@pedrobaeza , do you need a squash on this one ?

pedrobaeza:
Yes, at least the pylint one. And the other 2 requires some explanation on the commit message.

connector_prestashop/components/version_key.py
```diff
@@ -95,3 +95,48 @@ class VersionKey1616(Component):
...
+    _inherit = "_prestashop.version.key"
+    _usage = "prestashop.version.key.1.7.x.0"
+
+    keys = {
+        "product_option_value": "product_option_value",
+        "category": "categories",
+        "image": "image",
+        "order_slip": "order_slip",
+        "order_slip_detail": "order_slip_details",
+        "order_discounts": "order_cart_rules",
```
jesusVMayor, line 133: You need to change the _prestashop_name in prestashop.sale.order.line.discount.adapter model to:
```
@property
def _prestashop_model(self):
    return self.backend_record.get_version_ps_key("order_discounts")
```
An add the key in the other versions.

jesusVMayor (changes requested):
Works fine in 1.7 lower versions, now i tested it with 1.7.8.5 and there is a problem with 1.7.x keys

connector_prestashop/components/version_key.py
flotho: thanks @jesusVMayor I'll do this.

flotho:
> Works fine in 1.7 lower versions, now i tested it with 1.7.8.5 and there is a problem with 1.7.x keys

I've limited the version in the interface

flotho:
@pedrobaeza , 

> Yes, at least the pylint one. And the other 2 requires some explanation on the commit message.

Is it better for reading now ?

pedrobaeza:
I still see a waste of diff to first add something without linter and then pass the linter. Please apply directly on the first commit the linter for introducing the change at once.

flotho:
> I still see a waste of diff to first add something without linter and then pass the linter. Please apply directly on the first commit the linter for introducing the change at once.

@pedrobaeza thanks for your advice. Sorry to ask newbie question. I'm not sure what you're expecting. Is it a new rebase with all the precommit at the beginning ?

pedrobaeza:
More or less is:

Selección_028

or apply pre-commit on the content of each commit without adding an extra commit at the end.

jesusVMayor (approved):

pedrobaeza (changes requested):
Please avoid double diff without and with linter

pedrobaeza (approved):

pedrobaeza:
Thanks:

/ocabot merge major

flotho:
> Thanks:
> 
> /ocabot merge major

Thanks to you @pedrobaeza !
FMI, what is the best way to prevent this mixed. Do I need to start with a precommit on all files first and then deal with the functionnal changes? or is there another way to handle this ?

pedrobaeza:
Yes, if someone contributes without using linters, first of all is to "lint" the current content, and ideally put it in the same commit, as that diff doesn't bring any extra value (it's just linter), and then add on top what's needed.

PR #154, opened 2022-05-26
[POC][14.0] Reintroduce partner matching

jesusVMayor:
There is a matching_customer field since old versions without functionality. I am working on allow match by vat or email, but to make this change work, it's necessary to remove the sql_contraint prestashop_erp_uniq, because a partner could be duplicated on prestashop.
I don't see why this constraint is needed, if you don't configure the match, never will be necesarry, and if you want to match them, then you will prefeer to keep 2 prestashop records binded with an odoo record.
Or we can keep the constraint in particular models, but it's not necessary in all bindings.
What do you think?

### DIFF
--- a/connector_prestashop/migrations/14.0.2.0.0/pre-migrate.py
+++ b/connector_prestashop/migrations/14.0.2.0.0/pre-migrate.py
@@ -1,0 +1,18 @@
+# © 2022 Comunitea
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(cr, version):
+    cr.execute(
+        """
+            SELECT rel.relname, con.conname
+            FROM pg_catalog.pg_constraint con
+            INNER JOIN pg_catalog.pg_class rel
+                ON rel.oid = con.conrelid
+            INNER JOIN pg_catalog.pg_namespace nsp
+                ON nsp.oid = connamespace
+            WHERE con.conname ilike '%prestashop_erp_uniq%';
+        """
+    )
+    for constraint in cr.fetchall():
+        cr.execute(f"ALTER TABLE {constraint[0]} DROP CONSTRAINT {constraint[1]}")
--- a/connector_prestashop/models/binding/common.py
+++ b/connector_prestashop/models/binding/common.py
@@ -96,11 +96,3 @@ class PrestashopBindingOdoo(models.AbstractModel):
         string="Odoo binding",
         selection=_get_selection,
     )
-
-    _sql_constraints = [
-        (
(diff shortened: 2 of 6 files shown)


connector_prestashop/migrations/14.0.2.0.0/pre-migrate.py
```diff
@@ -0,0 +1,18 @@
+# © 2022 Comunitea
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+def migrate(cr, version):
```
flotho, line 5: I disagree with you.
This constrains is mandatory

connector_prestashop/models/res_partner/importer.py
```diff
@@ -32,6 +32,54 @@ class PartnerImportMapper(Component):
...
+    def odoo_id(self, record):
+        if not self.backend_record.matching_customer:
+            return {}
+        if self.backend_record.matching_customer_ch == "email":
+            partner_exists = self.env["res.partner"].search(
+                [
+                    ("email", "=", record.get("email")),
+                    "|",
+                    ("parent_id", "=", False),
+                    ("is_company", "=", True),
```
flotho, line 46: this one don't seems exact for me. 
We should consider that PS customer are by default is_company = False, this is why I also propose a PR in this direction : https://github.com/OCA/connector-prestashop/pull/162/files#diff-2ea12d5ea54d8b31980ec6332dc14982e121b5cfaffcafa76c99b9c760f28bfe

connector_prestashop/models/binding/common.py
```diff
@@ -96,11 +96,3 @@ def _get_selection(self):
         string="Odoo binding",
         selection=_get_selection,
     )
-
-    _sql_constraints = [
```
flotho, line 100: This deletion is not recommended.
First as it is in binding model, it's used everywhere in all binding so don't delete this one and moreover, it make sense to prevent Odoo having 2 binding with PS.
I also understand that in PS you could have twice the same mail and IMHO is a lack of consistency in PS. We're used to deal with this by asking to the sales team to identify which account is really active, archive the old one and change manually the mapping in Odoo. IMHO it's safer to deal manually the exceptions.

flotho (changes requested):
Hi @jesusVMayor a great thanks for your PR.
Actually I think that's a beahaviour I introduced my self in early v8/v10 versions.
Due to this experience and many uses cases faced with multiple customers, I submit you some changes request to improve your awesome feature.
Thanks in advance for the time spent to consider this message.
Regards

flotho:
successfully reintroduced in #155

PR #155, opened 2022-05-27
14.0 tax priority

flotho:
Require a rebase after https://github.com/OCA/connector-prestashop/pull/153 

Since 2021, in EU its mandatory to sell product with the VAT of the shipping country.

As a consequence some customers are expecting to have a single included tax price for example 120€TaxIncluded for all the countries.
PS don't support included tax as a cross country cross website reference for product price. So it's not possible in Ps to have a fixed price set with included vat and compute excluded tax price for each product and each country. As a consequence PS won't be able to compute a 100€w/oTax for france , 99€w/oTax for belgium etc...
Yet Odoo is able to do this thanks to fiscal positions and included VAT setup.
With this module, you can choose if Odoo is the master of the tax computation


florian-dacosta:
@flotho There is a big diff unrelated to the description of the PR, is it normal?

Anyway, regarding the new module, I don't really get it, if you let Odoo choosing the tax, the total amount will be different between prestashop and Odoo, right ?
Since the payment is done in prestashop, and there are also some exception rule to avoid the validation of sale orders iif odoo and prestashop amounts are different, I don't see how it helps.
Am I missing something important here ?

flotho:
> @flotho There is a big diff unrelated to the description of the PR, is it normal?
> 
Sure it is not ideal, but as the initial PR was not reviewed and as I need some part of this code I put everything in this PR.
I porpose to discuss about the main object and then I'll do a specific PR.
> Anyway, regarding the new module, I don't really get it, if you let Odoo choosing the tax, the total amount will be different between prestashop and Odoo, right ? Since the payment is done in prestashop, and there are also some exception rule to avoid the validation of sale orders iif odoo and prestashop amounts are different, I don't see how it helps. Am I missing something important here ?

This part is really tricky. You're totally right regarding the fact that the amounts in PS and in Odoo will be different.
Just imagine the following use case. A customer buy a product in france (original country of the company for our example) and it costs 120€ VAT included. with this case everything will be identical in PS and Odoo regarding the amount, taxe etc...

Then another customer buy the same product in Belgium and he will be charge for the same VAT INCLUDED amount because the seller don't want to change the price regarding the country. So the paid amount will be 120€VAT included. Yet, yet yet, in PS the VAT excluded amount won't be reduced by 21% but with 20% the french VAT. Why this, it's because the prices in PS are set up with a  "without VAT" base price, and if you change the setup in PS to use the tax of the country you'll have different tax included prices. 

With our setup, PS become only a "point fo sale" on internet for charging and getting payment whatever the country and always with the same amount.
Of course you'll have to deactivate the rules for amounts or at least adapt it to ignore foreign countries.

I hope this explanation is clear enough

connector_prestashop/models/product_product/importer.py
```diff
@@ -323,15 +323,25 @@ def odoo_id(self, record):
             code = record.get(self.backend_record.matching_product_ch)
             if self.backend_record.matching_product_ch == "reference":
                 if code:
-                    product = self.env["product.product"].search(
-                        [("default_code", "=", code)], limit=1
+                    product = (
+                        self.env["product.product"]
+                        .with_context(active_test=False)
```
flotho, line 328: FIX https://github.com/OCA/connector/issues/437

flotho:
Hi @florian-dacosta , I really would like to have your feedback regarding my explanation.  It's mostly based on my PS knowledge and you're probably able to propose another PS setup. From now, this is the simpliest way to handle such a case without "breaking" all the commercial work on PS part.

florian-dacosta:
Sorry for the delay, I was on holidays.
Well, your explanation is clear to me.
I don't like it very much, because the tax will be wrong (for foreign countries) in prestashop and so the tax displayed to the customer will be wrong... Also the fact to disable the exception rule about amounts is kind of dangerous I think and ccould lead to some errors.

But I have no better idea and I don't really know well prestashop, only the minimum to work with the odoo connector.
So I guess I could approve it (when green) since it is in a separated module, but I would not feel comfortable using it!

flotho:
> Sorry for the delay, I was on holidays. Well, your explanation is clear to me. I don't like it very much, because the tax will be wrong (for foreign countries) in prestashop and so the tax displayed to the customer will be wrong... Also the fact to disable the exception rule about amounts is kind of dangerous I think and ccould lead to some errors.
> 
> But I have no better idea and I don't really know well prestashop, only the minimum to work with the odoo connector. So I guess I could approve it (when green) since it is in a separated module, but I would not feel comfortable using it!

thanks for your feedback, I'll propose a new PR with additionnal exception rules so that this exception will be clear as a first instance

flotho:
closing because of too many changes , it will be split into different PR later on

PR #162, opened 2022-07-20
[IMP]Force partner type when PS company not set

flotho:
the company field in PS clearly indicates that it shoudl be a company in Odoo.
As a counterpart we should consider that if this field is not field, the PS account is a person.
This PR propose to force this value so that the general conf/behaviour in Odoo will not be the default behaviour for the connector.

### DIFF
--- a/connector_prestashop/models/res_partner/importer.py
+++ b/connector_prestashop/models/res_partner/importer.py
@@ -44,7 +44,7 @@ class PartnerImportMapper(Component):
     def is_company(self, record):
         if record.get("company"):
             return {"is_company": True}
-        return {}
+        return {"is_company": False}
 
     @mapping
     def birthday(self, record):


flotho:
Hi @jesusVMayor , as a complement of your PR, I would appreciate a reveiw on this one.
Regards

flotho:
Hi @saxomanu @gaelTorrecillas @rousseldenis , a small review on this would be appreciated.
regards

florian-dacosta (approved):
is_company default to False, so I think it does not change anything (I already have the expecting behavior without this fix) : https://github.com/OCA/OCB/blob/14.0/odoo/addons/base/models/res_partner.py#L210
But why not doing it explicitly

flotho:
Hi @florian-dacosta , you're totally right.
Except the fact that some additionnal default values could be added for customer having multiple activities like PSO and/or B2B. Forcing by code the default value for this "channel" is a way to preserve consistency of the datas coming from PS.

flotho:
ping @hailangvn @jesusVMayor any chance to be revbiewed ?

hailangvn (approved):
`Explicit is better than implicit` so this is good to me. Thanks for the work.

flotho:
ping @OCA/connector-prestashop-maintainers any chance to be merged ?

hailangvn (approved):

hailangvn (approved):
LGTM. Thanks.

bguillot:
Thank you:

/ocabot merge patch

PR #181, opened 2024-06-26
[16.0][MIG] connector_prestashop_catalog_manager: Migration to 16.0

AurelijaNorvaisaite:


cubells:
Migrated module that is needed:

https://github.com/OCA/product-attribute/pull/1967

